PR #7, opened 2021-03-26
[WIP, DO NOT MERGE] Rally study 01 support

rhelmer:
This closes issue 5, set up basic CI. We'll need to configure CircleCI and make sure it's ready before merging.


rhelmer:
@hamilton this PR merges your in-progress changes to WebScience, and only includes the modules we need for study-01. They all pass eslint, and I have stubs in for tests. Should we merge this as-is so we can get CircleCI going?

rhelmer:
Actually it looks like we're close enough to the `rally-prep` branch being usable and nice-enough-looking let's not worry about this for now. It'll be easy to remove currently-unused-by-study01 modules if we decide to.

Issue #24, opened 2021-03-28
Restore and refactor randomization module


jonathanmayer:
We temporarily removed this module because it wasn't used in the initial news and disinformation study, but it will be used in other studies. The module needs refactoring to match WebExtensions conventions.

https://github.com/mozilla-rally/web-science/blob/ion/Utilities/Randomization.js

PR #30, opened 2021-03-30
Refactored randomization module #24

jonathanmayer:


src/randomization.js
```diff
@@ -0,0 +1,107 @@
...
+            if(storageKey in retrievedConditions)
+                conditionCache = retrievedConditions[storageKey];
+            else
+                conditionCache = { };
+        }
+    }
+
+    // Try to load the selected condition from the cache
+    if(conditionSet.name in conditionCache)
+        return conditionCache[conditionSet.name];
```
rhelmer, line 78: I think always using braces is a better practice (someone always comes along and adds a condition without realizing) but let's do that with eslint and `eslint --fix` everything vs. bothering to do it here. I'll file an issue

rhelmer: (I mean it's not that terrible here because it's `return` or `throw`ing below, but if someone comes along and inserts a new line before the return things fail in new and interesting ways)

src/randomization.js
```diff
@@ -0,0 +1,107 @@
...
+        throw "The condition set must include an array with at least one condition."
+    for(const condition of conditionSet.conditions) {
+        if(condition.weight <= 0)
+            throw "Condition weights must be positive values."
+        totalWeight += condition.weight;
+        if(conditionNames.has(condition.name))
+            throw "Conditions must have unique names."
+        conditionNames.add(condition.name);
+    }
+    let randomValue = Math.random();
```
rhelmer, line 94: I went down the rabbit-hole of how `Math.random` works nowadays, I guess it's using https://en.wikipedia.org/wiki/Xoroshiro128%2B now, which sounds better than I remember!

src/webScience.js
```diff
@@ -53,3 +53,6 @@ export { linkExposure }
 
 import * as socialMediaLinkSharing from "./socialMediaLinkSharing.js"
 export { socialMediaLinkSharing }
+
+import * as randomization from "./randomization.js"
+export { randomization }
```
rhelmer, line 58: I wonder if we could do these as one line like: `export { randomization } from './randomization.js`? I haven't tried it yet but remember [seeing it on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export#re-exporting_aggregating).

jonathanmayer: I looked into this. There's a pending proposal for namespace re-exports: https://github.com/tc39/proposal-export-ns-from

rhelmer (approved):
I'll leave it to you whether JS `Math.random` is high-quality enough for the needs here, my knowledge about `Math.random` was [a little behind the times](https://v8.dev/blog/math-random) so that was interesting to learn about anyway.

src/webScience.js
rhelmer: Hm, wasn't that merged in https://github.com/tc39/ecma262/pull/1174 ? Although looking at the compatibility table on MDN we should probably avoid it for now anyway.

src/randomization.js
jonathanmayer: I thought about using the Web Crypto API for better randomness. I think we're OK with `Math.random` in this context, since the goal is just to divide users into a set of cohorts with target relative populations. Those cohorts will change (e.g., users withdrawing), and if researchers run statistical tests they'll have to account for cohort size.

src/webScience.js
jonathanmayer: Ah, I misunderstood the status of the proposal. Since we already need to use recent browser versions because of various WebExtensions APIs, I think we could switch to this syntax.

Issue #19, opened 2021-03-28
Standardize random ID generation


jonathanmayer:
We currently implement two different types of random ID generation:
* `pageManager.content.js`: 128-bit random hexadecimal strings.
* `userSurvey.js`:  Version 4 UUIDs, which are 122-bit random values with special formatting.
I don't have a strong preference between these options or others (e.g., random strings with a larger character set).

jonathanmayer:
Flagging that @rhelmer expressed a preference for using UUIDs and the [uuid](https://www.npmjs.com/package/uuid) library.

https://github.com/mozilla-rally/web-science/pull/37#pullrequestreview-629095211

PR #37, opened 2021-04-05
Refactoring the userSurvey module to remove study-specific code

### LINKED ISSUE #10: Replace study-specific code in UserSurvey
The current `UserSurvey` implementation has study-specific values hard-coded in several places. Those values should be passed into the module from study code.

jonathanmayer:
Closes #10. Getting the inlined survey content right was unexpectedly complicated...
  * A `browserAction` popup can load a page that's a `blob` URL but not a `data` URL.
  * The default WebExtensions CSP prevents loading inline scripts in that page, so we need to add `script-src` hashes of the scripts to study extension manifests.

Dealing with this nonsense makes me increasingly confident that the Rally core add-on should manage survey UX. See https://github.com/mozilla-rally/rally-core-addon/issues/177 and https://github.com/mozilla-rally/rally-core-addon/issues/178.


src/id.js
```diff
@@ -0,0 +1,19 @@
...
+ * Studies can use these identifiers to uniquely label events and other
+ * items of interest.
+ * @module webScience.id
+ */
+
+/**
+* Generate an ID, a random 128-bit value represented as a hexadecimal string.
+* @returns {string} The new random ID.
+*/
+export function generateId() {
```
rhelmer, line 12: This is probably OK but we should consider using [uuidjs](https://github.com/uuidjs/uuid) just because making sure it's RFC compliant is annoying. Using `window.crypto` is good though, it might be fine to keep it this way.

jonathanmayer: I've been waffling on whether to use UUIDs or something else (see https://github.com/mozilla-rally/web-science/issues/19). UUIDs are antiquated and inefficient... but they are common. 🤷‍♂️

src/userSurvey.js
```diff
@@ -1,118 +1,154 @@
...
+ *   * `'sha256-l+kgCjP15GJlVSDL9qMNffrHu8mxJcag42o2TYofOUM='`
  * @module webScience.userSurvey
  */
-import * as storage from "./storage.js"
-import * as messaging from "./messaging.js"
+import * as id from "./id.js";
+import * as storage from "./storage.js";
+import * as messaging from "./messaging.js";
+import * as inline from "./inline.js";
+import popupPromptPage from "./html/userSurvey.popupPrompt.html";
```
rhelmer, line 35: Huh, is it the rollup plugin using this? I've never seen HTML imported like this.

src/userSurvey.js
```diff
@@ -1,118 +1,154 @@
...
-            title: surveyRequestTitle,
-            iconUrl: SHIELD_URL
-        });
-        scheduleSurveyRequest(currentTime);
-    }
+/**
+ * Set a timeout to remind the user to complete the study.
+ */
+function scheduleReminderForUser() {
+    setTimeout(remindUser, Math.max((lastSurveyRequest + reminderInterval * millisecondsPerSecond) - Date.now(), 0));
```
rhelmer, line 81: When `setTimeout` fires, it might be polite to the user to have `remindUser` called by `window.requestIdleCallback` but maybe I'm overthinking it.

rhelmer (approved):
This lgtm overall, however we *really* need tests here as I am sure I will miss things just going over it by hand like this :) I'd suggest that we focus on unit tests first, although some things are going to be tough without an integration test I suspect (like the survey stuff)

src/id.js
rhelmer: > UUIDs are antiquated and inefficient... but they are common. 🤷‍♂️

Yeah a lot of that analysis tooling expects them :/ it's frustrating.

src/userSurvey.js
jonathanmayer: Yeah... I came up with this approach to inlining HTML. It's essentially the same as the approach to inlining content scripts.

src/userSurvey.js
jonathanmayer: I _think_ it should be fine... `remindUser` immediately yields the event loop with an `await`.

Issue #21, opened 2021-03-28
Clean up linkResolution


jonathanmayer:
We've known this particular module would benefit from a close review and cleanup for quite some time.

jonathanmayer:
A stale list of possible improvements: https://github.com/citp/news-disinformation-study/issues/60

PR #43, opened 2021-04-07
Lots of documentation fixes

jonathanmayer:
Plus a README section with module implementation status and a small API improvement in `linkResolution` #21.

### DIFF
--- a/src/matching.js
+++ b/src/matching.js
@@ -27,8 +27,7 @@ const matchPatternValidationRegExp = new RegExp("(^<all_urls>$)|(^(https?|wss?|f
 /**
  * A Set of URL schemes permitted in WebExtensions match patterns.
  * @see {@link https://searchfox.org/mozilla-central/source/toolkit/components/extensions/MatchPattern.cpp}
- * @constant
- * @type {Set<string>}
+ * @constant {Set<string>}
  * @private
  */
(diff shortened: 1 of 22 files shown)


src/matching.js
```diff
@@ -194,8 +209,9 @@ export function importMatchPatternSet(exportedMatchPatternSet) {
  * Future performance improvements could include:
  *   * Replacing the path matching implementation to eliminate regular expressions entirely.
  *   * Replacing the match pattern index, such as by implementing a trie.
+ * @private
  */
-class MatchPatternSet {
+class _MatchPatternSet {
```
rhelmer, line 214: Prefixing with `_` seems odd to me since this isn't done with the private functions elsewhere (since it's advisory anyway I tend to think something like `@private` is enough but I understand why people do this sort of thing in the absence of type checking).

Not exporting from the module seems like a pretty good indication that it's private to the module also?

rhelmer (approved):
I *really* like this table, I want to use this elsewhere! You can take or leave my comment about prefixing private things with `_` or not.

src/matching.js
jonathanmayer: The only reason I added that prefix was so that we could separately define a public `MatchPatternSet` type in JSDoc. Part of the broader problem of how we can better document class-like objects and events. 😕

PR #45, opened 2021-04-08
Remove code repetition from content scripts

### LINKED ISSUE #18: Remove code repetition from content scripts
Certain content scripts reuse code from background modules. Now that we're bundling dependencies into content scripts with Rollup, we can integrate that code via dependencies rather than copying and pasting code blocks into content scripts.

In `pageManager.content.js`:

- [x] Add an `events` module dependency (`Event` and `EventWithoutOptions`) and remove the copy and pasted `Event` and `EventWithoutOptions` classes.
- [x] Refactor the ID generation code into a background module and add the new module as a dependency.

In `linkExposure.content.js`:

- [x] Refactor the Facebook shim and AMP URL code into the `linkResolution` background module and add the module as a dependency.
- [x] Add a `matching` module dependency and remove the copy and pasted `MatchPatternSet` class.

In `socialMediaActivity.facebook.content.js`:

- [x] Add a `linkResolution` module dependency and remove the redundant code for handling Facebook shims and link decoration (see above).

jonathanmayer:
Resolves #18.

### DIFF
--- a/src/content-scripts/linkExposure.content.js
+++ b/src/content-scripts/linkExposure.content.js
@@ -3,115 +3,14 @@
  * @module webScience.linkExposure.content
  */
 
+import { importMatchPatternSet } from "../matching.js";
+import { parseFacebookLinkShim, parseAmpUrl } from "../linkResolution.js";
+
 // async IIFE wrapper to enable await syntax
 (async function () {
 
     let pageManager = null;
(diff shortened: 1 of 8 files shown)


rhelmer (approved):
So cool to be able to use modules here now.

PR #48, opened 2021-04-08
Adding basic WebExtensions permissions checking

### LINKED ISSUE #31: Document the WebExtensions permissions that each module requires and add support for permissions checking
We don't currently provide this information to library users, other than the permissions in `study-template`. We also don't provide an easy way for study developers to check that they've added all the required permissions.

jonathanmayer:
Resolves #31.

### DIFF
--- a/src/permissions.js
+++ b/src/permissions.js
@@ -1,0 +1,60 @@
+/**
+ * This module facilitates checking that required permissions are
+ * provided in the WebExtensions manifest.
+ * 
+ * @module webScience.permissions
+ */
+
+/**
+ * Check that the WebExtensions manifest includes specified API and origin permissions.
+ * @param {Object} options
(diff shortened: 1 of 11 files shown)


src/permissions.js
```diff
@@ -0,0 +1,60 @@
+/**
```
rhelmer, line 1: When I look at this I keep feeling like there must be a way to make it more concise, or that it should early-return, etc. but I think the fact that it is very specific about the warnings and gives a full report of problems is pretty nice to the caller. It feels repetitive but having one super generic function and 4 callers that will never be expanded doesn't seem like it adds any clarity.

rhelmer (approved):

src/permissions.js
rhelmer: Anyway, nothing to resolve here really just puzzled over it for a little bit, but I think it's good as-is. Nice work!

PR #49, opened 2021-04-08
Rewrite for linkResolution.resolveUrl

### LINKED ISSUE #21: Clean up linkResolution

jonathanmayer:
Resolves #21.

### DIFF
--- a/src/linkResolution.js
+++ b/src/linkResolution.js
@@ -14,6 +14,113 @@ permissions.check({
     suggestedOrigins: [ "<all_urls>" ]
 });
 
+// AMP caches and viewers
+
+/**
+ * A RegExp that matches and parses AMP cache and viewer URLs. If there is a match, the RegExp provides several
+ * named capture groups.
+ *   * AMP Cache Matches
+ *     * `ampCacheSubdomain` - The subdomain, which should be either a reformatted version of the
+ *       URL domain or a hash of the domain. If there is no subdomain, this capture group
+ *       is `undefined`.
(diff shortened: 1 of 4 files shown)


src/linkResolution.js
```diff
@@ -114,180 +99,365 @@ export function removeFacebookLinkDecoration(url) {
...
+    const urlObj = new URL(url);
+
+    // Twitter (t.co)
+    // Removing the amp=1 parameter results in more reliable HTTP redirects instead of HTML/JS redirects
+    if(urlObj.hostname === "t.co") {
+        urlObj.searchParams.delete("amp");
+    }
+
+    // Google News (news.google.com)
+    // Setting the User-Agent to curl results in more reliable HTTP redirects instead of HTML/JS redirects
```
rhelmer, line 286: huh, interesting.

rhelmer (approved):
OK, well I mean this seems reasonable in theory. This is an example of something that I think would benefit from an integration test since mocking the result of the API calls doesn't seem that useful, we have some [test code over in rally-core](https://github.com/mozilla-rally/rally-core-addon/blob/4818149d18bb7073f039046dca439ae63653ec42/tests/core-addon/integration/utils.js#L90) which allows generating little test extensions on-the-fly that would be useful for this sort of thing.

rhelmer:
>  lit

To be clear I am fine landing this as-is, I just wanted to make a note about testing this.

Issue #22, opened 2021-03-28
Optimize the timeouts in measurement content scripts


jonathanmayer:
Details: https://github.com/citp/news-disinformation-study/issues/77

- [x] In `pageNavigation.content.js`, measure page depth with a throttled `onscroll` listener (i.e., set a timer to check the scroll depth when `onscroll` fires and there isn't already a timer set).
- [x] In `linkExposure.content.js`, only tick the timer for measuring links when the page has user attention.

Issue #17, opened 2021-03-28
Complete event model migration for pageNavigation, linkExposure, socialMediaLinkSharing, and idle


jonathanmayer:
In the early versions WebScience, we implemented a measurement and storage model for these modules (i.e., each module ran an individual measurement and stored records of that measurement). We have partially transitioned to a WebExtensions-like event model for these modules, and there are some remaining items to complete.

Remaining items for `pageNavigation`:

- [x] Implement support for adding and removing multiple `pageDataCallback` listeners with differing `PageDataOptions` values.
- [x] Add logic to ensure that the associated content script only runs once per page.

Remaining items for `linkExposure`:

- [x] Implement support for adding and removing multiple `linkExposureCallback` listeners with differing `LinkExposureOptions` values.
- [x] Integrate untracked link exposures into `linkExposureCallback` and `LinkExposureDetails`, rather than using different (and undocumented) callback and detail types for untracked exposures. Note that the meaning of an untracked link will vary by listener, depending on its match patterns.
- [x] Add logic to ensure that the associated content script only runs once per page.

Remaining items for `socialMediaLinkSharing`:

- [ ] Implement support for adding and removing multiple `socialMediaShareCallback` listeners with differing `SocialMediaShareOptions` values.
- [ ] Correct the documentation for `SocialMediaShareOptions`.
- [ ] Add documentation for `SocialMediaShareDetails`.

Remaining items for `idle`:

- [x] Implement support for removing `onStateChanged` listeners.

jonathanmayer:
A stale list of items for `socialMediaLinkSharing`: https://github.com/citp/news-disinformation-study/issues/64

PR #50, opened 2021-04-10
Updates for linkExposure

jonathanmayer:
These updates simplify the content script logic, remove unused code, and complete the module's migration to WebExtensions event syntax #17. The changes also address the timer tick optimization in #22.


jonathanmayer:
@rhelmer I had a couple more ideas about how to improve the linkExposure module, so please don't bother reviewing this yet. I'll drop a note when the additional changes are done.

src/content-scripts/linkExposure.content.js
```diff
@@ -3,29 +3,43 @@
...
+    }
+
     let pageManager = null;
 
     /**
      * How often (in milliseconds) to check the page for new links.
      * @constant {number}
      */
-    const updateInterval = 2000;
+    const updateInterval = 3000;
```
rhelmer, line 30: We should consider making things like this configurable at runtime, not in this PR but something to think about.

src/linkExposure.js
```diff
@@ -21,246 +25,458 @@ permissions.check({
...
+    const resolvedLinkUrlPromises = linkUrls.map((linkUrl) => {
+        return linkResolution.resolveUrl(linkUrl);
+    });
+
+    // Once resolution is complete, notify the linkExposureUpdate listeners
+    Promise.allSettled(resolvedLinkUrlPromises).then(async (results) => {
+        // For each link URL, if we have a resolved URL, use that
+        // If we don't have a resolved URL, use the original URL with
+        // cache, shim, and link decoration parsing
+        for(let i = 0; i < linkUrls.length; i++) {
```
rhelmer, line 227: Nit, but if `linkUrls` is an array and you need the loop counter, I think you could use `for ... of` just to simplify things a little:

```suggestion
        for(const [i, linkUrl] of linkUrls.entries()) {
```

src/linkExposure.js
```diff
@@ -21,246 +25,458 @@ permissions.check({
...
+            if(results[i].status === "fulfilled") {
+                linkUrls[i] = results[i].value;
```
rhelmer, line 229:
```suggestion
                linkUrl = results[i].value;
```

src/linkExposure.js
```diff
@@ -21,246 +25,458 @@ permissions.check({
...
             }
             else {
-                await createLinkExposureRecord(linkExposure, nextLinkExposureIdCounter);
+                linkUrls[i] = await linkResolution.resolveUrl(linkUrls[i], { request: "none" });
```
rhelmer, line 232:
```suggestion
                linkUrl = await linkResolution.resolveUrl(linkUrl, { request: "none" });
```

src/linkExposure.js
rhelmer: Not important, just nice to use iterators and not worry about off-by-one etc. (of course in this case since you're indexing it into this other `results` array it doesn't help /that/ much :)) Makes it more consistent with the use of `Map` elsewhere etc

src/linkExposure.js
```diff
@@ -21,246 +25,458 @@ permissions.check({
...
-            nonmatchingLinkExposures: "number",
-            linkExposures: "object"
+        // If we are ignoring self links, determine whether each link URL is a self link
+        // by comparing to the page URL's public suffix + 1
+        // These are links that do not appear to be self links in the content
+        // script, but resolve to self links
+        let selfLinks = null;
+        if(ignoreSelfLinks) {
+            const pagePS1 = linkResolution.urlToPS1(url);
+            selfLinks = linkUrls.map(linkUrl => { return pagePS1 === linkResolution.urlToPS1(linkUrl); })
```
rhelmer, line 243: Nit but I think you can just use the implicit return of arrow functions ("concise body"):

```suggestion
            selfLinks = linkUrls.map(linkUrl => pagePS1 === linkResolution.urlToPS1(linkUrl))
```

src/linkExposure.js
```diff
@@ -21,246 +25,458 @@ permissions.check({
...
 
-    initialized = true;
+        // Notify the listeners
+        for(const [listener, listenerRecord] of linkExposureUpdateListeners) {
+            // Check private window and page match pattern requirements for the listener
+            if((!privateWindow || listenerRecord.privateWindows) &&
+            listenerRecord.pageMatchPatternSet.matches(url)) {
+                const matchingLinkUrls = [];
+                let nonmatchingLinkCount = 0;
+                for(let i = 0; i < linkUrls.length; i++) {
```
rhelmer, line 253: Same thing as above w/ `for ... of` and `linkUrls.entries()` iterator, if you want.

src/linkResolution.js
```diff
@@ -44,6 +45,16 @@ export const ampRegExp = new RegExp(
...
     , "i");
 
+/**
+ * A MatchPatternSet for AMP caches and viewers.
+ * @constant {matching.MatchPatternSet}
+ */
+export const ampMatchPatternSet = matching.createMatchPatternSet(
+    matching.domainsToMatchPatterns(ampCacheDomains, false).concat(
+        ampViewerDomainsAndPaths.map((ampViewerDomainAndPath) => {
+            return `*://${ampViewerDomainAndPath}*`;
```
rhelmer, line 55: You could do implicit return here too if you think it makes it more readable.

src/pageManager.js
```diff
@@ -184,6 +184,8 @@ const considerUserInputForAttention = true;
  * @constant {Events.Event<pageVisitStartListener, PageVisitStartListenerOptions>}
  */
 export const onPageVisitStart = events.createEvent({
+    // Make sure the module is initialized when a listener is added
+    addListenerCallback: listener => { initialize(); },
```
rhelmer, line 188:
```suggestion
    addListenerCallback: listener => initialize(),
```

src/pageManager.js
```diff
@@ -230,6 +232,8 @@ function pageVisitStart(details) {
  * @constant {Events.Event<pageVisitStopListener, PageVisitStartListenerOptions>}
  */
 export const onPageVisitStop = events.createEvent({
+    // Make sure the module is initialized when a listener is added
+    addListenerCallback: listener => { initialize(); },
```
rhelmer, line 236:
```suggestion
    addListenerCallback: listener => initialize(),
```

rhelmer (approved):
OK I only have the nit-pickiest nits, feel free to ignore them if you like, choose whatever you think is more readable.

src/content-scripts/linkExposure.content.js
jonathanmayer: Definitely. What I'd like us to support, eventually, is two sets of configurable `linkExposure` options: one that's per-listener (e.g., link and page match patterns) and one that's for all listeners (e.g., timer intervals and self-links).

PR #53, opened 2021-04-10
Functionality for series of subsequent surveys and updated browser action popup UI

### LINKED ISSUE #55: Add functionality to userSurvey.js to allow for series of subsequent surveys
The current implementation of userSurvey.js only allows for a single survey per study. For my search engine study, there is an initial study at the start of the study as well as a follow-up survey after 30 days. The questions in the follow-up survey depend on if the user completed the initial survey.

To allow for subsequent surveys, userSurvey.js will need the following features:

- A `surveyName` property in the `options` parameter of `setSurvey` that identifies the currently active survey and is saved in storage.
- A `getCurrentSurveyName` method that returns the name of the current survey .
- A `getSurveyCompletionStatus` method that returns the completion status of the current survey.
- `setSurvey` clears the values of `lastSurveyRequest`, `surveyCompleted`, and `surveyCancelled` when the `surveyName` passed into `setSurvey` does not match the `surveyName` in storage.

kartkand:
Fixes #55 and #59 

New browser action popups:

Available survey with icon
Screen Shot 2021-04-19 at 11 27 00 AM

Available survey without icon
Screen Shot 2021-04-19 at 11 26 13 AM

No survey with icon
Screen Shot 2021-04-19 at 11 22 48 AM

No survey without icon
Screen Shot 2021-04-19 at 11 23 43 AM

### DIFF
--- a/src/html/userSurvey.popupNoPrompt.html
+++ b/src/html/userSurvey.popupNoPrompt.html
@@ -4,6 +4,7 @@
         <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
         <meta content="utf-8" http-equiv="encoding">
         <style>
+            /* May need to add font bundling or revisit the font choice for cross-browser compatibility. */
             body {
                 font-family: Inter, X-LocaleSpecific, sans-serif;
             }
--- a/src/html/userSurvey.popupNoPrompt.html
+++ b/src/html/userSurvey.popupNoPrompt.html
@@ -3,18 +3,42 @@
     <head>
         <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
         <meta content="utf-8" http-equiv="encoding">
+        <style>
+            body {
+                font-family: Inter, X-LocaleSpecific, sans-serif;
+            }
+
+            .icon {
+                max-width: 50px;
+                max-height: 50px;
+                float: left;
+                padding-left: 10px;
+                padding-right: 10px; 
+            }
+        </style>
     </head>
     <body>
-        <div id="content" class="paragraph" style="width: 300px; padding-right: 25px; padding-left: 25px;">
+        <div style="display: flex; max-width: 300px; padding-top: 10px; padding-bottom: 10px;">
+            <img id="icon" class="icon" style="display: none;">
+            <div id="content" class="paragraph" style="padding-left: 10px; padding-right: 10px;"></div>
         </div>
         <script>
             (async function() {
                 // Load the message from storage
-                const storageKey = "webScience.userSurvey.popupNoPromptMessage";
-                const contentResults = await browser.storage.local.get(storageKey);
-                if(storageKey in contentResults) {
+                const messageStorageKey = "webScience.userSurvey.popupNoPromptMessage";
+                const messageResults = await browser.storage.local.get(messageStorageKey);
+                if(messageStorageKey in messageResults) {
                     const contentElement = document.getElementById("content");
-                    contentElement.innerHTML = contentResults[storageKey];
+                    contentElement.innerHTML = messageResults[messageStorageKey];
+                }
+
+                // Load icon URL from storage
+                const iconUrlStorageKey = "webScience.userSurvey.iconUrl";
+                const iconUrlResults = await browser.storage.local.get(iconUrlStorageKey);
+                if(iconUrlStorageKey in iconUrlResults) {
+                    const iconElement = document.getElementById("icon");
+                    iconElement.src = iconUrlResults[iconUrlStorageKey];
+                    iconElement.style.display = "block";
                 }
             })();
         </script>
--- a/src/html/userSurvey.popupNoPrompt.html
+++ b/src/html/userSurvey.popupNoPrompt.html
@@ -29,7 +29,7 @@
                 const messageResults = await browser.storage.local.get(messageStorageKey);
                 if(messageResults[messageStorageKey]) {
                     const contentElement = document.getElementById("content");
-                    contentElement.innerHTML = messageResults[messageStorageKey];
+                    contentElement.textContent = messageResults[messageStorageKey];
                 }
 
                 // Load icon URL from storage
--- a/src/html/userSurvey.popupNoPrompt.html
+++ b/src/html/userSurvey.popupNoPrompt.html
@@ -6,20 +6,19 @@
         <style>
             /* May need to add font bundling or revisit the font choice for cross-browser compatibility. */
             body {
-                font-family: Inter, X-LocaleSpecific, sans-serif;
+                font-family: sans-serif;
             }
 
             .icon {
                 max-width: 50px;
                 max-height: 50px;
                 float: left;
-                padding-left: 10px;
                 padding-right: 10px; 
             }
         </style>
     </head>
     <body>
-        <div style="display: flex; max-width: 300px; padding-top: 10px; padding-bottom: 10px;">
+        <div style="display: flex; max-width: 280px; padding: 10px;">
             <img id="icon" class="icon" style="display: none;">
             <div id="content" class="paragraph" style="padding-left: 10px; padding-right: 10px;"></div>
         </div>
@@ -28,15 +27,15 @@
                 // Load the message from storage
                 const messageStorageKey = "webScience.userSurvey.popupNoPromptMessage";
                 const messageResults = await browser.storage.local.get(messageStorageKey);
-                if(messageStorageKey in messageResults) {
+                if(messageResults[messageStorageKey]) {
                     const contentElement = document.getElementById("content");
                     contentElement.innerHTML = messageResults[messageStorageKey];
                 }
 
                 // Load icon URL from storage
-                const iconUrlStorageKey = "webScience.userSurvey.iconUrl";
+                const iconUrlStorageKey = "webScience.userSurvey.popupIconUrl";
                 const iconUrlResults = await browser.storage.local.get(iconUrlStorageKey);
-                if(iconUrlStorageKey in iconUrlResults) {
+                if(iconUrlResults[iconUrlStorageKey]) {
                     const iconElement = document.getElementById("icon");
                     iconElement.src = iconUrlResults[iconUrlStorageKey];
                     iconElement.style.display = "block";
--- a/src/html/userSurvey.popupPrompt.html
+++ b/src/html/userSurvey.popupPrompt.html
@@ -83,7 +83,7 @@
                 const messageResults = await browser.storage.local.get(messageStorageKey);
                 if(messageResults[messageStorageKey]) {
                     const contentElement = document.getElementById("content");
-                    contentElement.innerHTML = messageResults[messageStorageKey];
+                    contentElement.textContent = messageResults[messageStorageKey];
                 }
 
                 // Load icon URL from storage
--- a/src/html/userSurvey.popupPrompt.html
+++ b/src/html/userSurvey.popupPrompt.html
@@ -4,6 +4,7 @@
         <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
         <meta content="utf-8" http-equiv="encoding">
         <style>
+            /* May need to add font bundling or revisit the font choice for cross-browser compatibility. */
             body {
                 font-family: Inter, X-LocaleSpecific, sans-serif;
             }
--- a/src/html/userSurvey.popupPrompt.html
+++ b/src/html/userSurvey.popupPrompt.html
@@ -4,19 +4,22 @@
         <meta content="text/html;charset=utf-8" http-equiv="Content-Type">
         <meta content="utf-8" http-equiv="encoding">
         <style>
+            body {
+                font-family: Inter, X-LocaleSpecific, sans-serif;
+            }
+
             .button-container {
-                width: 100%;
-                height: 50px;
                 display: table;
                 border-collapse: separate;
                 border-spacing: 0px;
             }
 
             .buttonBlue {
-                background-color: #4cb1f5;
-                width: 100px;
-                color: black;
-                padding: 15px;
+                background-color: #145EE3;
+                width: 99px;
+                color: white;
+                padding-top: 14px;
+                padding-bottom: 14px;
                 text-align: center;
                 text-decoration: none;
                 font-size: 16px;
@@ -25,14 +28,15 @@
             }
 
             .buttonBlue:hover {
-                background-color: #31a6f5;
+                background-color: #104FBF;
             }
 
             .buttonGray {
-                background-color: #d9d9d9;
+                background-color: #f1f1f1;
                 width: 100px;
                 color: black;
-                padding: 15px;
+                padding-top: 14px;
+                padding-bottom: 14px;
                 text-align: center;
                 text-decoration: none;
                 font-size: 16px;
@@ -41,26 +45,54 @@
             }
 
             .buttonGray:hover {
-                background-color: #b5b5b5;
+                background-color: #e2e2e2;
+            }
+
+            .icon {
+                max-width: 50px;
+                max-height: 50px;
+                float: left;
+                padding-left: 10px;
+                padding-right: 10px; 
+            }
+
+            .separator {
+                width: 1px;
+                height: 46px;
+                color: black;
+                float: left;
+                background: #d3d3d3;
             }
         </style>
     </head>
     <body>
-        <div id="content" class="paragraph" style="width: 300px; padding-right: 25px; padding-left: 25px;">
+        <div style="display: flex; max-width: 300px; padding-top: 10px; padding-bottom: 10px;">
+            <img id="icon" class="icon" style="display: none;">
+            <div id="content" class="paragraph" style="padding-left: 10px; padding-right: 10px;"></div>
         </div>
         <div class="button-container">
-            <a href="#" class="buttonGray" name="later">Later</a>
-            <a href="#" class="buttonGray" name="never">Never</a>
-            <a href="#" class="buttonBlue" name="agree">Continue</a>
+            <a href="#" class="button buttonGray" name="later">Later</a>
+            <span class="separator"></span>
+            <a href="#" class="button buttonGray" name="never">Never</a>
+            <a href="#" class="button buttonBlue" name="agree">Continue</a>
         </div>
         <script>
             (async function() {
                 // Load the message from storage
-                const storageKey = "webScience.userSurvey.popupPromptMessage";
-                const contentResults = await browser.storage.local.get(storageKey);
-                if(storageKey in contentResults) {
+                const messageStorageKey = "webScience.userSurvey.popupPromptMessage";
+                const messageResults = await browser.storage.local.get(messageStorageKey);
+                if(messageStorageKey in messageResults) {
                     const contentElement = document.getElementById("content");
-                    contentElement.innerHTML = contentResults[storageKey];
+                    contentElement.innerHTML = messageResults[messageStorageKey];
+                }
+
+                // Load icon URL from storage
+                const iconUrlStorageKey = "webScience.userSurvey.iconUrl";
+                const iconUrlResults = await browser.storage.local.get(iconUrlStorageKey);
+                if(iconUrlStorageKey in iconUrlResults) {
+                    const iconElement = document.getElementById("icon");
+                    iconElement.src = iconUrlResults[iconUrlStorageKey];
+                    iconElement.style.display = "block";
                 }
                 
                 // Listen for clicks on the buttons
--- a/src/html/userSurvey.popupPrompt.html
+++ b/src/html/userSurvey.popupPrompt.html
@@ -6,7 +6,7 @@
         <style>
             /* May need to add font bundling or revisit the font choice for cross-browser compatibility. */
             body {
-                font-family: Inter, X-LocaleSpecific, sans-serif;
+                font-family: sans-serif;
             }
 
             .button-container {
@@ -53,8 +53,7 @@
                 max-width: 50px;
                 max-height: 50px;
                 float: left;
-                padding-left: 10px;
-                padding-right: 10px; 
+                padding-right: 10px;
             }
 
             .separator {
@@ -67,7 +66,7 @@
         </style>
     </head>
     <body>
-        <div style="display: flex; max-width: 300px; padding-top: 10px; padding-bottom: 10px;">
+        <div style="display: flex; max-width: 280px; padding: 10px;">
             <img id="icon" class="icon" style="display: none;">
             <div id="content" class="paragraph" style="padding-left: 10px; padding-right: 10px;"></div>
         </div>
@@ -82,15 +81,15 @@
                 // Load the message from storage
                 const messageStorageKey = "webScience.userSurvey.popupPromptMessage";
                 const messageResults = await browser.storage.local.get(messageStorageKey);
-                if(messageStorageKey in messageResults) {
+                if(messageResults[messageStorageKey]) {
                     const contentElement = document.getElementById("content");
                     contentElement.innerHTML = messageResults[messageStorageKey];
                 }
 
                 // Load icon URL from storage
-                const iconUrlStorageKey = "webScience.userSurvey.iconUrl";
+                const iconUrlStorageKey = "webScience.userSurvey.popupIconUrl";
                 const iconUrlResults = await browser.storage.local.get(iconUrlStorageKey);
-                if(iconUrlStorageKey in iconUrlResults) {
+                if(iconUrlResults[iconUrlStorageKey]) {
                     const iconElement = document.getElementById("icon");
                     iconElement.src = iconUrlResults[iconUrlStorageKey];
                     iconElement.style.display = "block";
--- a/src/userSurvey.js
+++ b/src/userSurvey.js
@@ -27,8 +27,8 @@
  * require special Content Security Policy permissions in the extension
  * manifest (the `"content_security_policy"` key). Those permissions
  * are currently the following additional `script-src` values.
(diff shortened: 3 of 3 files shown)


src/userSurvey.js
```diff
@@ -171,7 +196,6 @@ export async function setSurvey(options) {
     let lastSurveyRequest = await storageSpace.get("lastSurveyRequest");
     const surveyCompleted = await storageSpace.get("surveyCompleted");
     const surveyCancelled = await storageSpace.get("surveyCancelled");
-    lastSurveyRequest = await storageSpace.get("lastSurveyRequest");
```
jonathanmayer, line 174: Don't we need this? Otherwise calling `setSurvey` will always pop a tab open.

src/userSurvey.js
```diff
@@ -222,5 +246,25 @@ export async function setSurvey(options) {
...
  */
 export async function getSurveyId() {
+    initializeStorage();
     return await storageSpace.get("surveyId");
 }
+
+/**
+ * Gets the completion status of the current survey. Can be used if a
+ * subsequent survey depends on the status of the previous survey. 
+ * @returns {boolean} - Whether the current survey has been completed.
```
jonathanmayer, line 256: What if there is no current survey?

src/userSurvey.js
```diff
@@ -222,5 +246,25 @@ export async function setSurvey(options) {
...
+ */
+export async function getSurveyCompletionStatus() {
+    initializeStorage();
+    return await storageSpace.get("surveyCompleted");
+}
+
+/**
+ * Gets the name of the current survey.
+ * @returns {string} - The name of the current survey.
```
jonathanmayer, line 265: What if there is no current survey?

src/userSurvey.js
```diff
@@ -162,7 +173,21 @@ export async function setSurvey(options) {
...
+    let currentSurvey = await storageSpace.get("currentSurvey");
+
+    /* Clears the existing survey data if this survey differs from the
+     * previous survey.
+     */
+    if (currentSurvey !== options.surveyName) {
+        await storageSpace.set("lastSurveyRequest", null);
+        await storageSpace.set("surveyCompleted", false);
+        await storageSpace.set("surveyCancelled", false);
+    }
```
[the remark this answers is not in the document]
jonathanmayer: Also, the current implementation seems bugged... `createdSurvey` will be `true` after the first call to `setSurvey`. A subsequent call will result in an error because of that. There would have to be an extension reload between calls of `setSurvey`.

src/userSurvey.js
```diff
@@ -185,7 +209,7 @@ export async function setSurvey(options) {
     }
 
     // If this is the first survey request, open the survey in a new tab
-    if (lastSurveyRequest === null) {
+    if(lastSurveyRequest === null) {
```
jonathanmayer, line 212: We don't want to generate a new survey ID for each survey. We want one survey ID for the duration of the study.

jonathanmayer (commented):
Added comments. This still needs some work. I'm going to untag @rhelmer for review until we're closer.

src/userSurvey.js
kartkand: We're already setting the value of `lastSurveyRequest` from storage 3 lines above this.

src/userSurvey.js
kartkand: I moved the functionality of generating a new survey ID into the `getSurveyId` method. This method now generates a survey ID if there is no value for survey ID in storage.

src/userSurvey.js
kartkand: Added to the description of the function saying that the function returns null if there is no current survey.

src/userSurvey.js
kartkand: Added to the description of the function saying that the function returns null if there is no current survey.

src/userSurvey.js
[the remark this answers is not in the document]
kartkand: Created `endCurrentSurvey` function that:

1. Clears the existing survey reminder timeout, removes the listener
2. Removes the `browser.webRequest.onBeforeRequest` listener that checks for survey completion
3. Clears the values of `lastSurveyRequest`, `surveyCompleted`, `surveyCancelled`, and `currentSurvey` in storage.

src/userSurvey.js
```diff
@@ -49,12 +49,19 @@ permissions.check({
 let storageSpace = null;
 
 /**
- * Whether a survey has already been created. This module
- * currently only supports one survey per study.
+ * The ID of the survey reminder timeout (is null if there 
+ * is no such timeout).
+ * @type {number}
```
jonathanmayer, line 54:
```suggestion
 * @type {number|null}
```

src/userSurvey.js
```diff
@@ -122,9 +129,30 @@ function setPopupToNoPromptPage() {
...
+/**
+ * Called when the current survey is completed. Sets surveyCompleted to true in storage
+ * and changes the browser action popup to the survey's no prompt page.
+ * @private
+ */
+function setSurveyComplete() {
+    storageSpace.set("surveyCompleted", true);
+    setPopupToNoPromptPage();
+}
+
```
jonathanmayer, line 151: Would use a different function name, to avoid confusion with the similar `setSurvey`.

src/userSurvey.js
```diff
@@ -144,11 +172,6 @@ function setPopupToNoPromptPage() {
  * platform (e.g., SurveyMonkey, Typeform, Qualtrics, etc.).
  */
 export async function setSurvey(options) {
-    if(createdSurvey) {
-        throw new Error("userSurvey only supports one survey at present.");
-    }
-    createdSurvey = true;
-
```
jonathanmayer, line 151: Proposed logic here:
* If there's a call to `setSurvey` with the same survey name as the survey in storage, ignore the function parameters, load all the survey attributes from storage, and continue with the survey. That allows studies with just one survey to have one call to `setSurvey` and be done.
* If there's a call to `setSurvey` with a different survey name from the survey in storage, throw an error, because two surveys are now set at the same time.
* If there's a call to `setSurvey` and there's no survey in storage, save the parameters in storage and carry out the survey.

src/userSurvey.js
```diff
@@ -190,37 +214,89 @@ export async function setSurvey(options) {
...
 
-    // Set listeners for cancel and open survey button clicks in the survey request
-    messaging.onMessage.addListener(() => {
-        storageSpace.set("surveyCancelled", true);
-        setPopupToNoPromptPage();
-    }, { type: "webScience.userSurvey.cancelSurvey" });
-    messaging.onMessage.addListener(openSurveyInNewTab, { type: "webScience.userSurvey.openSurvey" });
+    // Listeners for cancel and open survey button click only need to be added once.
+    // They do not need to be added again for subsequent calls to setSurvey.
+    if (listenersRegistered === false) {
```
jonathanmayer, line 231:
```suggestion
    if (!listenersRegistered) {
```

src/userSurvey.js
```diff
@@ -190,37 +214,89 @@ export async function setSurvey(options) {
...
+    listenersRegistered = true;
 }
 
 /**
  * Each study participant has a persistent survey ID, generated with
  * the id module. The ID is automatically added as a parameter to
  * the survey URL, enabling researchers to import survey data from an
- * external platform and sync it with Rally data.
+ * external platform and sync it with Rally data. This method returns the
+ * survey ID (generating it if it does not already exist)
```
jonathanmayer, line 248:
```suggestion
 * survey ID, generating it if it does not already exist.
```

src/userSurvey.js
```diff
@@ -190,37 +214,89 @@ export async function setSurvey(options) {
...
+        surveyId = id.generateId();
+        await storageSpace.set("surveyId", surveyId);
+    }
+    return surveyId;
+}
+
+/**
+ * Gets the completion status of the current survey. Can be used if a
+ * subsequent survey depends on the status of the previous survey.
+ * @returns {boolean} - Whether the current survey has been completed.
```
jonathanmayer, line 264:
```suggestion
 * @returns {boolean|null} - Whether the current survey has been completed.
```

src/userSurvey.js
```diff
@@ -190,37 +214,89 @@ export async function setSurvey(options) {
...
+ * Returns null if there is no current survey.
+ */
+export async function getSurveyCompletionStatus() {
+    initializeStorage();
+    return await storageSpace.get("surveyCompleted");
+}
+
+/**
+ * Gets the name of the current survey.
+ * @returns {string} - The name of the current survey. Returns null
```
jonathanmayer, line 274:
```suggestion
 * @returns {string|null} - The name of the current survey. Returns null
```

src/userSurvey.js
```diff
@@ -190,37 +214,89 @@ export async function setSurvey(options) {
...
+    initializeStorage();
+    return await storageSpace.get("currentSurvey");
 }
+
+/**
+ * End the current survey. Should be called before a subsequent survey is started.
+ */
+ export async function endCurrentSurvey() {
+    // If there is an existing survey reminder timeout, clears the timeout.
+    if (reminderTimeoutId !== null) {
```
jonathanmayer, line 287: Don't need this check, `clearTimeout` will (by specification) just do nothing with an invalid timeout ID.

jonathanmayer (commented):
This is looking good. My only big-picture comment is about how `setSurvey` should behave in various circumstances.

src/userSurvey.js
kartkand: I implemented this, but slightly confused by what your first bullet meant. Studies will still need to call `setSurvey` upon every restart of the extension regardless of if survey attributes can be loaded from storage.

src/userSurvey.js
```diff
@@ -49,12 +49,19 @@ permissions.check({
...
+
+/**
+ * Whether listeners for this module have already been registered.
  * @type {boolean}
  * @private
  */
-let createdSurvey = false;
+let listenersRegistered = false;
 
 // Module-wide variables for a survey, set in createSurvey
```
jonathanmayer, line 66: Would be good to provide documentation for each of the module-wide variables. Would also make sense to match the object type used as a parameter for `setSurvey` and in storage.

src/userSurvey.js
```diff
@@ -122,9 +129,30 @@ function setPopupToNoPromptPage() {
...
+ * and changes the browser action popup to the survey's no prompt page.
+ * @private
+ */
+function completeSurvey() {
+    storageSpace.set("surveyCompleted", true);
+    setPopupToNoPromptPage();
+}
+
 /**
  * Prompt the user to respond to a survey.
```
jonathanmayer, line 153: Would give a short description of what the function does (i.e., how it behaves in various circumstances), an example of how to use it when you're running just one survey in your study, and an example of how to use it when you're running two surveys.

src/userSurvey.js
```diff
@@ -144,34 +172,50 @@ function setPopupToNoPromptPage() {
...
-    /* Check when we last asked the user to do the survey. If it's null,
-     * we've never asked, which means the extension just got installed.
-     * Open a tab with the survey, and save this time as the most recent
-     * request for participation.
-     */
+    // Check when we last asked the user to do the survey. If it's null,
+    // we've never asked, which means the extension just got installed.
+    // Open a tab with the survey, and save this time as the most recent
+    // request for participation.
     let lastSurveyRequest = await storageSpace.get("lastSurveyRequest");
```
jonathanmayer, line 216:
```suggestion
    lastSurveyRequest = await storageSpace.get("lastSurveyRequest");
```
This is a bug I introduced. `lastSurveyRequest` should not be redeclared in function scope.

src/userSurvey.js
```diff
@@ -190,37 +234,88 @@ export async function setSurvey(options) {
...
     );
 
-    // Set listeners for cancel and open survey button clicks in the survey request
-    messaging.onMessage.addListener(() => {
-        storageSpace.set("surveyCancelled", true);
-        setPopupToNoPromptPage();
-    }, { type: "webScience.userSurvey.cancelSurvey" });
-    messaging.onMessage.addListener(openSurveyInNewTab, { type: "webScience.userSurvey.openSurvey" });
+    // Listeners for cancel and open survey button click only need to be added once.
+    // They do not need to be added again for subsequent calls to setSurvey.
```
jonathanmayer, line 250: Would either remove these listeners in `endCurrentSurvey` or explain why it's OK to leave them in place.

src/userSurvey.js
```diff
@@ -15,7 +15,7 @@
  * 
  * # Limitations
  * Note that this module is currently very limited: it only supports
- * one survey per study, with few options and a constrained design.
+ * one survey at a time per study, with few options and a constrained design.
```
jonathanmayer, line 18: "the user complete the survey" → "the user to complete the survey" above. Would also point the reader to the  the documentation for the exported functions.

src/userSurvey.js
```diff
@@ -190,37 +234,88 @@ export async function setSurvey(options) {
...
+    initializeStorage();
+    return await storageSpace.get("surveyCompleted");
+}
+
+/**
+ * Gets the name of the current survey.
+ * @returns {string|null} - The name of the current survey. Returns null
+ * if there is no current survey.
+ */
+export async function getCurrentSurveyName() {
```
jonathanmayer, line 297: Would be consistent about using `Current` in survey function names. I think it isn't necessary (e.g., `setSurvey` and `getSurveyCompletionStatus`).

src/userSurvey.js
```diff
@@ -190,37 +234,88 @@ export async function setSurvey(options) {
...
+ * Gets the completion status of the current survey. Can be used if a
+ * subsequent survey depends on the status of the previous survey.
+ * @returns {boolean|null} - Whether the current survey has been completed.
+ * Returns null if there is no current survey.
+ */
+export async function getSurveyCompletionStatus() {
+    initializeStorage();
+    return await storageSpace.get("surveyCompleted");
+}
+
```
jonathanmayer, line 291: Would generalize this beyond completion, since that use case is foreseeable. Something like `getSurveyStatus` that returns null if there is no survey, "active" if there is a survey, "cancelled" if the user cancelled the survey, or "completed" if the user completed the survey.

src/userSurvey.js
```diff
@@ -190,37 +234,88 @@ export async function setSurvey(options) {
...
+    return surveyId;
 }
+
+/**
+ * Gets the completion status of the current survey. Can be used if a
+ * subsequent survey depends on the status of the previous survey.
+ * @returns {boolean|null} - Whether the current survey has been completed.
+ * Returns null if there is no current survey.
+ */
+export async function getSurveyCompletionStatus() {
```
jonathanmayer, line 287: This doesn't need to be `async`, the module could cache the survey status in module-wide variables.

src/userSurvey.js
jonathanmayer: Also doesn't need to be `async`, for the same reason as above.

jonathanmayer (commented):
Almost there. Just some minor remaining requests and then over to @rhelmer.

src/userSurvey.js
kartkand: I think keeping this as `async` would be beneficial because a developer could immediately make a call to this function without having to initialize the module.

kartkand: Keeping this as `async` so that no prior initialization needs to occur before calling this function.

src/userSurvey.js
kartkand: Keeping this as async so that no prior initialization needs to occur before calling this function.

src/userSurvey.js
kartkand: Added documentation but not matching the object type used as a parameter for `setSurvey` because `lastSurveyRequest` is not passed in as a parameter and some of the properties of the passed in `options` parameter do not need to be available module-wide.

src/userSurvey.js
```diff
@@ -190,37 +295,106 @@ export async function setSurvey(options) {
...
+    // They do not need to be added again for subsequent calls to setSurvey.
+    // These listeners do not need to be removed in endCurrentSurvey because they will
+    // not receive messages when the popup is the no prompt page.
+    if (!listenersRegistered) {
+        // Set listeners for cancel and open survey button clicks in the survey request.
+        messaging.onMessage.addListener(() => {
+            storageSpace.set("surveyCancelled", true);
+            setPopupToNoPromptPage();
+        }, { type: "webScience.userSurvey.cancelSurvey" });
+        messaging.onMessage.addListener(openSurveyInNewTab, { type: "webScience.userSurvey.openSurvey" });
```
jonathanmayer, line 320: Should either make `openSurveyInNewTab` synchronous or wrap this call in an anonymous synchronous function. When you have an `async` message listener (as I put in, which was a bug), that listener always returns a `Promise`, which can cause problems for other listeners. See: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onMessage

src/userSurvey.js
```diff
@@ -190,37 +295,106 @@ export async function setSurvey(options) {
...
-    messaging.onMessage.addListener(openSurveyInNewTab, { type: "webScience.userSurvey.openSurvey" });
+    // Listeners for cancel and open survey button click only need to be added once.
+    // They do not need to be added again for subsequent calls to setSurvey.
+    // These listeners do not need to be removed in endCurrentSurvey because they will
+    // not receive messages when the popup is the no prompt page.
+    if (!listenersRegistered) {
+        // Set listeners for cancel and open survey button clicks in the survey request.
+        messaging.onMessage.addListener(() => {
+            storageSpace.set("surveyCancelled", true);
+            setPopupToNoPromptPage();
```
jonathanmayer, line 318: Need to remove the survey completion URL listener if the user canceled the survey.

src/userSurvey.js
```diff
@@ -123,55 +199,84 @@ function setPopupToNoPromptPage() {
...
-    surveyUrl = options.surveyUrl;
-    reminderIconUrl = browser.runtime.getURL(options.reminderIcon);
-    reminderInterval = options.reminderInterval;
-    reminderTitle = options.reminderTitle;
-    reminderMessage = options.reminderMessage;
+    surveyUrl = surveyDetails.surveyUrl;
+    reminderIconUrl = browser.runtime.getURL(surveyDetails.reminderIcon);
+    reminderInterval = surveyDetails.reminderInterval;
+    reminderTitle = surveyDetails.reminderTitle;
+    reminderMessage = surveyDetails.reminderMessage;
```
jonathanmayer, line 260: Could condense this with destructuring assignment.

jonathanmayer (commented):
Great work on the survey API, and thanks for helping improve the popup appearance. Please update #59 with screenshots, but don't close the issue since we'll probably want to do more work on the popup. Once you're done with that and the remaining items, please tag in @rhelmer for a last review.

kartkand:
@rhelmer could you take a look at this PR and let me know if you have any feedback?

src/html/userSurvey.popupNoPrompt.html
```diff
@@ -3,18 +3,42 @@
...
+            }
+        </style>
     </head>
     <body>
-        <div id="content" class="paragraph" style="width: 300px; padding-right: 25px; padding-left: 25px;">
+        <div style="display: flex; max-width: 280px; padding: 10px;">
+            <img id="icon" class="icon" style="display: none;">
+            <div id="content" class="paragraph" style="padding-left: 10px; padding-right: 10px;"></div>
         </div>
         <script>
```
rhelmer, line 25: Hm, I'm surprised that the default CSP rules allow inline script:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Content_Security_Policy#default_content_security_policy

src/html/userSurvey.popupPrompt.html
```diff
@@ -41,26 +46,53 @@
...
                 // Load the message from storage
-                const storageKey = "webScience.userSurvey.popupPromptMessage";
-                const contentResults = await browser.storage.local.get(storageKey);
-                if(storageKey in contentResults) {
+                const messageStorageKey = "webScience.userSurvey.popupPromptMessage";
+                const messageResults = await browser.storage.local.get(messageStorageKey);
+                if(messageResults[messageStorageKey]) {
                     const contentElement = document.getElementById("content");
-                    contentElement.innerHTML = contentResults[storageKey];
+                    contentElement.innerHTML = messageResults[messageStorageKey];
```
rhelmer, line 86: `innerHTML` might be OK, but it may lead to additional scrutiny. `createElement` et al. might be better.

src/userSurvey.js
```diff
@@ -184,43 +291,118 @@ export async function setSurvey(options) {
...
+    clearTimeout(reminderTimeoutId);
+
+    // Remove any previously added listener for browser.webRequest.onBeforeRequest
+    // that checks for the survey completion URL.
+    browser.webRequest.onBeforeRequest.removeListener(surveyCompletionUrlListener);
+
+    initializeStorage();
+
+    // Clears the the data in storage for the current survey.
+    await storageSpace.set("lastSurveyRequest", null);
```
rhelmer, line 404: Hm, just wondering why `[clear()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage/StorageArea/clear)` is not exposed from the storage module?

rhelmer (changes requested):
I'd like to test this a little bit (we really need to have a test extension be built and run as part of CI!), one thing I was surprised to find that works is to run `web-ext lint` in the root, which seems to work even if there is no `manifest.json`:

```
% web-ext lint                                   
Applying config file: ./package.json
Validation Summary:

errors          1              
notices         0              
warnings        8              

ERRORS:

Code                    Message                       Description                                                                         File   Line   Column
TYPE_NO_MANIFEST_JSON   manifest.json was not found   No manifest.json was found at the root of the extension. The package file must be                       
                                                      a ZIP of the extension's files themselves, not of the containing directory. See:                        
                                                      https://mzl.la/2r2McKv for more on packaging.                                                           
WARNINGS:

Code                    Message                             Description                                                                         File                                        Line   Column
INLINE_SCRIPT           Inline scripts blocked by default   Default CSP rules prevent inline JavaScript from running                            intermediate/html/userSurvey.popupNoProm…                
                                                            (https://mzl.la/2pn32nd).                                                           pt.html                                                  
INLINE_SCRIPT           Inline scripts blocked by default   Default CSP rules prevent inline JavaScript from running                            intermediate/html/userSurvey.popupPrompt…                
                                                            (https://mzl.la/2pn32nd).                                                           .html                                                    
INLINE_SCRIPT           Inline scripts blocked by default   Default CSP rules prevent inline JavaScript from running                            src/html/userSurvey.popupNoPrompt.html                   
                                                            (https://mzl.la/2pn32nd).                                                                                                                    
INLINE_SCRIPT           Inline scripts blocked by default   Default CSP rules prevent inline JavaScript from running                            src/html/userSurvey.popupPrompt.html                     
                                                            (https://mzl.la/2pn32nd).                                                                                                                    
UNSAFE_VAR_ASSIGNMENT   Unsafe assignment to innerHTML      Due to both security and performance concerns, this may not be set using dynamic    intermediate/content-scripts/pageContent…   1279   10    
                                                            values which have not been adequately sanitized. This can lead to security issues   .content.js                                              
                                                            or fairly serious performance degradation.                                                                                                   
UNSAFE_VAR_ASSIGNMENT   Unsafe assignment to innerHTML      Due to both security and performance concerns, this may not be set using dynamic    intermediate/content-scripts/pageContent…   1586   8     
                                                            values which have not been adequately sanitized. This can lead to security issues   .content.js                                              
                                                            or fairly serious performance degradation.                                                                                                   
UNSAFE_VAR_ASSIGNMENT   Unsafe assignment to innerHTML      Due to both security and performance concerns, this may not be set using dynamic    intermediate/content-scripts/pageText.co…   1279   10    
                                                            values which have not been adequately sanitized. This can lead to security issues   ntent.js                                                 
                                                            or fairly serious performance degradation.                                                                                                   
UNSAFE_VAR_ASSIGNMENT   Unsafe assignment to innerHTML      Due to both security and performance concerns, this may not be set using dynamic    intermediate/content-scripts/pageText.co…   1586   8     
                                                            values which have not been adequately sanitized. This can lead to security issues   ntent.js                                                 
                                                            or fairly serious performance degradation.       
```

Ignoring the files that are not part of this patch, I suspect that when the CSP warnings here are resolved we'll also get warnings about `innerHTML` assignment. As long as the input is santizied this is usually OK, but `createElement` can be safer, because you can better separate the HTML generation parts (`createElement`) from the text content (`.textContent` can't be used to generate HTML, for instance)

src/html/userSurvey.popupNoPrompt.html
jonathanmayer: They don't! We use script hashes for these popup inline scripts.

src/userSurvey.js
jonathanmayer: We don't currently store all the keys associated with a `KeyValueStorage` storage space (which is just a key namespace subset of extension storage), so there isn't a way to clear the storage space without iterating all the keys in extension storage. We could add a `clear` function (and a `remove` function), those just haven't been necessary yet.

src/html/userSurvey.popupPrompt.html
jonathanmayer: How about `innerText`?

rhelmer: Yes! If all you need to modify is text, then this is a great way to [prevent XSS attacks](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innerhtml) in case unsanitized user input is ever accidentally passed here.

kartkand: Great, I switched this to use `textContent`.

Issue #23, opened 2021-03-28
Improvements to socialMediaActivity


jonathanmayer:
A couple stale lists of items:
* https://github.com/citp/news-disinformation-study/issues/64
* https://github.com/citp/news-disinformation-study/issues/54

PR #54, opened 2021-04-10
Update social media modules

akohlbre:
- Adds support for blocking to the events module (to allow for a module to wait on the result of a listener)
- Updates `socialMediaActivity` to use the event model
- Updates `socialMediaLinkSharing` to use the new interface to `socialMediaActivity`
- Adds documentation in `socialMediaActivity`

Remaining tasks for these modules are in #17 and #23.


src/events.js
```diff
@@ -119,15 +119,19 @@ class Event {
...
+                if((this.notifyListenersCallback === null) || this.notifyListenersCallback(listener, listenerArguments, options)) {
+                    const listenerReturnValue = listener.apply(null, listenerArguments);
+                    if (listenerReturnValue !== undefined) { returnVal = listenerReturnValue; }
+                }
             }
             catch(error) {
                 debugLog(`Error in listener notification: ${error}`);
             }
         });
+        return returnVal;
```
jonathanmayer, line 139: Rather than clobbering the return value for each listener, so the `notifyListeners` caller only receives the last return value, what about storing all the return values in an array and returning that to the `notifyListeners` caller?

src/socialMediaActivity.js
```diff
@@ -65,82 +66,61 @@ function registerPlatformListener(platform, eventType, blockingType, callback) {
...
-    }
-    if (events.includes("favorite") || events.includes("<all_events>")) {
-        registerPlatformListener("twitter", "favorite", blocking ? "blocking" : "nonblocking", callback);
+function addListener(listener, { platform, eventTypes, blocking=false }){
+    for (const eventType of eventTypes) {
+        registerPlatformListener(platform, eventType, blocking ? "blocking" : "nonblocking");
     }
-    tweetContentInit();
+    if (platform == 'twitter') tweetContentInit();
+    if (platform == 'facebook') fbPostContentInit();
```
jonathanmayer, line 102: === seems right here.

src/socialMediaActivity.js
```diff
@@ -65,82 +66,61 @@ function registerPlatformListener(platform, eventType, blockingType, callback) {
...
 
 /**
- * Register a callback for specific Facebook events. Supported events are "post", "reshare",
- * "react" (includes like, love, etc), and "comment".
- * @param callback - the function to call when the event happens
- * @param [String] events - array of events to be tracked
- * @param blocking - whether the listener should be blocking. Allows canceling the event.
+ * Filter generated events to only the listeners for the specific event type.
+ * @param {socialMediaActivityCallback} listener - The listener that may be notified.
+ * @param {Array} listenerArguments - The event about to be sent to the listener.
```
jonathanmayer, line 108: Is there a reason to use an array here instead of just the argument?

src/socialMediaActivity.js
```diff
@@ -65,82 +66,61 @@ function registerPlatformListener(platform, eventType, blockingType, callback) {
...
-        registerPlatformListener("facebook", "react", blocking ? "blocking" : "nonblocking", callback);
-    }
-    if (events.includes("comment") || events.includes("<all_events>")) {
-        registerPlatformListener("facebook", "comment", blocking ? "blocking" : "nonblocking", callback);
-    }
-    fbPostContentInit();
+function notifyFilter(listener, listenerArguments, options) {
+    const reportedEvent = listenerArguments[0];
+    const ret = reportedEvent['platform'] == options['platform'] &&
+        options['eventTypes'].includes(reportedEvent['eventType']);
```
jonathanmayer, line 115: `===` and `""`?

src/socialMediaActivity.js
```diff
@@ -65,82 +66,61 @@ function registerPlatformListener(platform, eventType, blockingType, callback) {
...
+ * A callback function for the social media activity event.
+ * @callback socialMediaActivityCallback
+ * @param {SocialMediaShareDetails} details - Additional information about the social media activity event.
+ */
+
+/**
+ * Options when adding a social media activity event listener.
+ * @typedef {Object} socialMediaActivityOptions
+ * @property {string} platform - The platform to add a listener for ('facebook', 'twitter', or 'reddit')
+ * @property {Array<string>} eventTypes - The events on the platform to notify for. For facebook, can
```
jonathanmayer, line 129: I've been trying to standardize on `string[]` for string array types.

src/socialMediaActivity.js
```diff
@@ -65,82 +66,61 @@ function registerPlatformListener(platform, eventType, blockingType, callback) {
...
-        registerPlatformListener("reddit", "postVote", blocking ? "blocking" : "nonblocking", callback);
-    }
-    if (events.includes("commentVote") || events.includes("<all_events>")) {
-        registerPlatformListener("reddit", "commentVote", blocking ? "blocking" : "nonblocking", callback);
-    }
-}
+
+/**
+ * @type {events.event<socialMediaActivityCallback, socialMediaActivityOptions}
+ */
```
jonathanmayer, line 137: See how events are annotated in other modules. We might eventually change the format, but it's the least-bad I could come up with for now.

src/socialMediaActivity.js
```diff
@@ -65,82 +66,61 @@ function registerPlatformListener(platform, eventType, blockingType, callback) {
...
+export const onSocialMediaActivity = events.createEvent({
+    addListenerCallback: addListener,
+    notifyListenersCallback: notifyFilter
+});
```
jonathanmayer, line 141: A big-picture design question: do we want one event for all the platforms, or one event per platform? The advantage of the former is we avoid some repetition. The advantage of the latter is library users might shoot themselves in the foot by mixing and matching platforms and activity types.

src/socialMediaActivity.js
```diff
@@ -30,15 +31,15 @@ export function enablePrivateWindows() {
     privateWindows = true;
 }
 
-/** Unregister old handlers for an event, and register a new one, if necessary.
+/**
+ * Unregister old handlers for an event, and register a new one, if necessary.
  * Unregistering is only necessary when there's already a nonblocking handler registered
  * and we want to convert it to a blocking handler.
  * @param platform - which social media platform the event is for
  * @param eventType - which type of event we're registering
```
jonathanmayer, line 56: Perhaps we should use the term "activity type" rather than "event type", since that's consistent with the name of the module and events are a specific concept in WebExtensions.

src/socialMediaActivity.js
```diff
@@ -82,82 +84,61 @@ function registerPlatformListener(platform, eventType, blockingType, callback) {
...
-    }
-    if (events.includes("commentVote") || events.includes("<all_events>")) {
-        registerPlatformListener("reddit", "commentVote", blocking ? "blocking" : "nonblocking", callback);
-    }
-}
+
+/**
+ * Options when adding a social media activity event listener.
+ * @typedef {Object} socialMediaActivityOptions
+ * @property {string} platform - The platform to add a listener for ('facebook', 'twitter', or 'reddit')
```
jonathanmayer, line 128: `""` and .

src/socialMediaActivity.js
```diff
@@ -82,82 +84,61 @@ function registerPlatformListener(platform, eventType, blockingType, callback) {
...
-        registerPlatformListener("reddit", "postVote", blocking ? "blocking" : "nonblocking", callback);
-    }
-    if (events.includes("commentVote") || events.includes("<all_events>")) {
-        registerPlatformListener("reddit", "commentVote", blocking ? "blocking" : "nonblocking", callback);
-    }
-}
+
+/**
+ * Options when adding a social media activity event listener.
+ * @typedef {Object} socialMediaActivityOptions
```
jonathanmayer, line 127: I've been trying to establish a convention of lower camel case for callback types and upper camel case for other types (including `@callback` functions that aren't really callbacks).

src/socialMediaActivity.js
```diff
@@ -776,6 +732,20 @@ function checkFacebookPostAudience(requestDetails) {
 }
 
 /**
+ * Given an object and the name of a target field, attempt to find a field with that name
+ * within the object. This function is necessary to handle the frequest small changes that
+ * Facebook makes to its APIs. For example, they sometimes change a field from being a
+ * string to being an array of strings that only ever holds a single string, or vice versa.
+ * The function checks recurses down the object structure, unwinding fields as it goes:
```
jonathanmayer, line 739: "checks recurses" = ? And "unwinding fields" = ?

src/socialMediaActivity.js
```diff
@@ -776,6 +732,20 @@ function checkFacebookPostAudience(requestDetails) {
 }
 
 /**
+ * Given an object and the name of a target field, attempt to find a field with that name
```
jonathanmayer, line 735: What does "field" mean?

src/socialMediaActivity.js
```diff
@@ -733,7 +685,11 @@ function verifyFacebookComment({requestDetails = null}) {
 
     return {};
 }
+
 /**
+ * Determine the audience of a Facebook post or reshare.
```
jonathanmayer, line 690: Here and throughout, when parsing platform-specific values, please make sure to document what the source data is and what assumptions you're making in parsing.

src/socialMediaActivity.js
```diff
@@ -852,6 +822,11 @@ async function extractFacebookReshare({requestDetails = null, verified = null, e
 }
 
 /**
+ * Ask our Facebook content script whether the last reshare was from a page or a person.
+ * This information is not included in the web request, nor is it attached to the generated post,
+ * so the only way to get it is by watching the user click the reshare button, which
+ * the content script does.
+ * @return {Promise} - The source of the last reshare ("person" or "page").
```
jonathanmayer, line 829: What type of Promise?

src/socialMediaActivity.js
```diff
@@ -1113,6 +1092,11 @@ function extractRedditCommentVote({requestDetails = null, eventTime = null}) {
     });
 }
 
+/**
+ * For a given subreddit name, return whether it is private or public.
+ * @param {string} subredditName - The name of the subreddit, without leading "r/".
+ * @return {Promise} - resolves to a string representing the subreddit's status, or "unknown".
```
jonathanmayer, line 1098: What type of Promise?

akohlbre:
This PR now contains all the code changes we're currently planning. I'll be pushing more documentation updates shortly, namely:
* Longer explanatory comments about usage at the top of SMA and SMLS
* Using the `@namespace` convention for documenting events and their associated object
* Clearer documentation for which fields are relevant to which event types in SMA
* More specific documentation about how we interact with platforms
* More specific documentation for a few return types (e.g. anything currently just `@returns {Object}`)

src/socialMediaActivity.js
```diff
@@ -777,6 +946,26 @@ function checkFacebookPostAudience(requestDetails) {
...
+ *  a regular expression to parse that field of the object instead is a good idea.
+ * @param {Object} object - The parent object to search.
+ * @param {string} fieldName - The target field to find.
+ * @param {boolean} enterArray - whether to use the first element of a found array as
+ *  the return value, or the entire array.
+ * @param {integer} recurseLevel - Used internally to prevent recursing down very large structures.
+ * @param - the value of the fieldName, if found, or null.
  * @private
  */
 function findFieldFacebook(object, fieldName, enterArray = true, recurseLevel = 5) {
```
jonathanmayer, line 971: Would give a more descriptive name.

src/socialMediaActivity.js
```diff
@@ -822,46 +1011,69 @@ function findFieldFacebook(object, fieldName, enterArray = true, recurseLevel =
...
+ * the content script does.
+ * @returns {string} - The source of the last reshare ("person" or "page").
  * @private
  */
-async function getReshareInfo() {
+function getReshareInfo() {
     return browser.tabs.sendMessage(facebookTabId, {"recentReshare": true}).then((response) => {
         return response;
     }, (e) => { console.log("ERROR", e); } );
 }
```
jonathanmayer, line 1070: Would use the `messaging` module and a schema. Would also give a more descriptive error.

src/socialMediaActivity.js
```diff
@@ -822,46 +1011,69 @@ function findFieldFacebook(object, fieldName, enterArray = true, recurseLevel =
...
+ * Ask our Facebook content script whether the last reshare was from a page or a person.
+ * This information is not included in the web request, nor is it attached to the generated post,
+ * so the only way to get it is by watching the user click the reshare button, which
+ * the content script does.
+ * @returns {string} - The source of the last reshare ("person" or "page").
  * @private
  */
-async function getReshareInfo() {
+function getReshareInfo() {
     return browser.tabs.sendMessage(facebookTabId, {"recentReshare": true}).then((response) => {
```
jonathanmayer, line 1067: This logic seems busted if the user has multiple Facebook pages open. Might be better to just send a message to the background script when there's a reshare and then consume the reshare event when notifying listeners.

src/socialMediaActivity.js
```diff
@@ -931,10 +1143,19 @@ function verifyFacebookReshare({requestDetails = null }) {
...
  * Get the contents and attachments of a Facebook post.
- * @param postId - the unique ID of the post
- * @param ownerId - the unique ID of the owner, or of the group, if the post is in a group
+ * Note: Do not pass Facebook post IDs as normal numbers, as they are too large to
+ *  maintain precision. Use strings instead.
+ * @param {string} postId - the unique ID of the post
+ * @returns {facebookPostContentsDetails}
  */
 export function getFacebookPostContents(postId) {
     return new Promise((resolve, reject) => {
```
jonathanmayer, line 1161: Same comment as the similar functionality for Twitter. This seems like it could be handled in the background page.

src/socialMediaActivity.js
```diff
@@ -949,79 +1170,167 @@ export function getFacebookPostContents(postId) {
 
 /**
  * Reddit posts don't currently have validation needs.
+ * @param {Object} requestDetails - the raw request.
+ * @returns {Object} - an empty object.
  * @private
  */
 function verifyRedditPost({requestDetails = null}) {
     return {};
 }
```
jonathanmayer, line 1179: Seems fine to omit this.

src/socialMediaActivity.js
```diff
@@ -949,79 +1170,167 @@ export function getFacebookPostContents(postId) {
...
+ * @returns {redditActivityDetails} - default-value Reddit details.
+ */
+function createBaseRedditObject() {
+    return {
+        eventType: "", eventTime: 0, postText: "", postAttachments: [],
+        subredditName: "", postTitle: "", voteDirection: 0, votedCommentId: "",
+        commentAttachments: [], commentText: "", commentedPostId: "", votedPostId: ""
+    };
+}
+
```
jonathanmayer, line 1192: See comments about similar functions for Twitter and Facebook.

src/socialMediaActivity.js
```diff
@@ -949,79 +1170,167 @@ export function getFacebookPostContents(postId) {
...
+    try {
         const postUrl = requestDetails.requestBody.formData["url"][0];
-        details.attachment = postUrl;
+        details.postAttachments.push(postUrl);
+    } catch { //ignore
     }
 
-    details.postTitle = requestDetails.requestBody.formData.title[0];
-    details.eventType = "post";
+    let richtext_json = null;
```
jonathanmayer, line 1226: See comment above about underscore variable names.

src/socialMediaActivity.js
```diff
@@ -1094,26 +1431,31 @@ function verifyRedditCommentVote({requestDetails = null}) {
...
 }
 
+/**
+ * For a given subreddit name, return whether it is private or public.
+ * @param {string} subredditName - The name of the subreddit, without leading "r/".
+ * @returns {Promise} - resolves to a string representing the subreddit's status, or "unknown".
+ */
 export function checkSubredditStatus(subredditName) {
     if (subredditName == "") return "unknown";
     return new Promise((resolve, reject) => {
```
jonathanmayer, line 1461: Here and throughout, you don't need to wrap `fetch` in a new `Promise`, because it already starts a `Promise` chain.

src/socialMediaActivity.js
```diff
@@ -1139,17 +1481,49 @@ export function checkSubredditStatus(subredditName) {
     });
 }
 
+function addIfExists(sourceObject, destinationObject, sourceFieldName,
```
jonathanmayer, line 1484: Documentation!

src/socialMediaLinkSharing.js
```diff
@@ -11,40 +11,78 @@ import * as socialMediaActivity from "./socialMediaActivity.js";
...
  * Options when adding a social media share event listener.
- * @typedef {Object} SocialMediaShareOptions
- * @property {string[]} [matchPattern=[]] - The webpages of interest for the measurement, specified with WebExtensions match patterns.
+ * @typedef {Object} socialMediaShareOptions
+ * @property {string[]} [destinationMatchPatterns=[]] - The webpages of interest for the
+ *  measurement, specified with WebExtensions match patterns.
+ * @property {boolean} facebook - whether to measure shares that happen on Facebook.
+ * @property {boolean} twitter - whether to measure shares that happen on Twitter.
+ * @property {boolean} reddit - whether to measure shares that happen on Reddit.
+ */
```
jonathanmayer, line 59: See comments on Slack.

src/socialMediaLinkSharing.js
```diff
@@ -11,40 +11,78 @@ import * as socialMediaActivity from "./socialMediaActivity.js";
...
+const debugLog = debugging.getDebuggingLog("socialMediaLinkSharing");
+
+/**
+ * Unlike posts on Facebook, which have individual privacy controls, or posts on Reddit, whose
+ * privacy is determined by the subreddit in which they are posted, the privacy of a Tweet is
+ * determined by a global setting for each account. We pull this setting once and store it.
+ * @type {string}
+ * @private
+ */
+let twitterPrivacySetting = "unknown";
```
jonathanmayer, line 76: This seems like it should go in `socialMediaActivity`.

src/socialMediaLinkSharing.js
```diff
@@ -60,13 +98,16 @@ function removeListener(listener) {
...
-    name: "webScience.socialMediaLinkSharing.onShare",
-    addListenerCallback: addListener,
-    removeListenerCallback: removeListener
-});
+function createBaseShareObject() {
+    return {
+        platform: "", shareTime: 0, url: "", audience: "", eventType: "", reshareSource: "",
+        type: "", untrackedCount: 0
+    };
+}
```
jonathanmayer, line 110: See other comments about these base object functions.

src/socialMediaLinkSharing.js
```diff
@@ -84,16 +125,76 @@ async function startMeasurement({
     reddit = false
 }) {
```
jonathanmayer, line 126: Should replace the start and stop measurement conventions, since we don't use those anymore.

src/socialMediaLinkSharing.js
```diff
@@ -84,16 +125,76 @@ async function startMeasurement({
...
     if (reddit) {
-        socialMediaActivity.registerRedditActivityTracker(redditLinks, ["post"]);
+        socialMediaActivity.onRedditActivity.addListener(redditLinks, {
+            eventTypes: ['post']});
     }
     if (twitter) {
-        socialMediaActivity.registerTwitterActivityTracker(twitterLinks, ["tweet", "retweet", "favorite"]);
+        socialMediaActivity.onTwitterActivity.addListener(twitterLinks, {
+            eventTypes: ['tweet', 'retweet', 'favorite']});
     }
```
jonathanmayer, line 138: Seems like some logic is needed in here for adding multiple listeners.

src/socialMediaLinkSharing.js
```diff
@@ -84,16 +125,76 @@ async function startMeasurement({
...
+}
+
+/**
+ * Given events that were set up to be "tracked"-type notifications but do not match the current
+ *  listener's match patterns, turn them into one or more "untracked"-type notifications instead.
+ * @param {socialMediaShareDetails[]} trackedEvents - a set of non-matching share events.
+ * @returns {socialMediaShareDetails[]} - the tracked events stripped and condensed into untracked events.
+ * @private
+ */
+function trackedToUntracked(trackedEvents) {
```
jonathanmayer, line 174: Would just combine this with the prior function.

src/socialMediaLinkSharing.js
```diff
@@ -84,16 +125,76 @@ async function startMeasurement({
...
+                tracked: false,
+                untrackedCount: 1
+            };
+            newEvents.push(newEvent);
+        }
+    }
+    return newEvents;
 }
 
 /**
```
jonathanmayer, line 200: Need to add logic for teardown after removing listeners.

src/socialMediaLinkSharing.js
```diff
@@ -104,62 +205,73 @@ function stopMeasurement() {
...
+ * Quote tweets are implemented as regular tweets that contain a link to another tweet.
+ * This function determines whether a given link is to a tweet or not.
+ * @param {string} url - a link to check.
+ * @returns {string[]} - result of matching regex on url.
  * @private
  */
 function isTwitterLink(url) {
-    const twitterLink = /twitter\.com\/[0-9|a-z|A-Z|_]*\/status\/([0-9]*)\//;
+    const twitterLink = /twitter\.com\/[0-9|a-z|A-Z|_]*\/status\/([0-9]*)/;
```
jonathanmayer, line 215: Would parse to a `URL`, check the hostname, then check the path.

src/socialMediaLinkSharing.js
```diff
@@ -11,40 +11,78 @@ import * as socialMediaActivity from "./socialMediaActivity.js";
 import * as linkResolution from "./linkResolution.js";
 
 /**
- * @constant {debugging.debuggingLogger}
- * @private
+ * @type {Events.Event<socialMediaShareCallback, socialMediaShareOptions>}
  */
-const debugLog = debugging.getDebuggingLog("socialMediaLinkSharing");
+export const onShare = events.createEvent({
```
jonathanmayer, line 16: I think the boundary between `socialMediaActivity` and `socialMediaLinkSharing` should be a bit different from what it is. Currently, `socialMediaActivity` exposes functions for looking up tweets, posts, and subreddits, and `socialMediaLinkSharing` has to call those functions. It would be preferable if we could keep those functions internal to `socialMediaActivity` and only expose them if study developers eventually need them. The information that `socialMediaLinkSharing` has to obtain itself is also information that many callers to `socialMediaActivity` would want anyway. A possible approach would be a listener option for `socialMediaActivity` that indicates shared tweets and posts should be resolved, plus adding subreddit status to reddit events.

src/socialMediaActivity.js
```diff
@@ -28,17 +29,155 @@ permissions.check({
...
+
+
+/**
+ * An object providing details about an event on Twitter.
+ * @typedef {Object} twitterActivityDetails
+ * @property {string} eventType - action taken by user ("tweet", "retweet", or "favorite").
+ * @property {number} eventTime - timestamp of user action.
+ * @property {string} tweetText - for a tweet, the content of the tweet.
+ * @property {string[]} tweetAttachments - for a tweet, any links attached to the tweet.
+ * @property {string} sharedId - for a favorite or share, the Twitter ID of the favorited or retweeted tweet.
```
jonathanmayer, line 83: Why not include a post ID for all Twitter, Facebook, and Reddit events?

src/socialMediaActivity.js
```diff
@@ -28,17 +29,155 @@ permissions.check({
  */
```
jonathanmayer, line 29: The `permissions.check` call above should go in the add listener callback functions, with the right match pattern and permissions (`webRequestBlocking`?) for the event type and listener options.

PR #60, opened 2021-04-15
Support for Content Security Policy checking

### LINKED ISSUE #57: Minimize and check special Content Security Policy requirements
Firefox currently exempts `moz-extension://` URLs from CSP, but not `blob:moz-extension://` URLs (see [bug 1294996](https://bugzilla.mozilla.org/show_bug.cgi?id=1294996)). That's a good idea, for the same reasons that CSP by default blocks blob URLs from a page's origin (i.e., an attacker might trick the page into executing code via a blob URL just like it could trick the page into executing code via `eval`). But that does create problems for us, since we want to load inlined JavaScript for the browser action popup (i.e., `userSurvey`) and for Web Workers (e.g., `pageText`).

Some specific items to address:
- [x] Attempt to integrate the CSP requirements for the `userSurvey` browser action popup into the HTML for the popup. If that's doable, we won't need CSP permissions for popup scripts in the extension manifest.
- [x] Get script hash permissions working for worker scripts (e.g., `pageText`), so we don't require a blanket blob URL permission in CSP. That's not a good security practice, and it might be a blocker for AMO review. From the MDN [documentation](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_security_policy) on WebExtensions CSP: "extensions with 'unsafe-eval', 'unsafe-inline', remote script, blob, or remote sources in their CSP are not allowed for extensions listed on addons.mozilla.org due to major security issues."
- [x] Add CSP checking to the permissions module so we can warn study developers when they're missing CSP requirements in the extension manifest.

jonathanmayer:
Plus a Content Security Policy check in `userSurvey`. Resolves #57, for now.

### DIFF
--- a/src/permissions.js
+++ b/src/permissions.js
@@ -6,12 +6,104 @@
  */
 
 /**
+ * An object where keys are Content Security Policy directive names and values are arrays of directive values.
+ * @typedef {Object} ContentSecurityPolicy
+ * @example
+ * {
+ *   "script-src": [ "'self'", "www.example.com"],
+ *   "object-src": [ 'self' ]
+ * }
+ */
+
+/**
+ * The Content Security Policy directives specified in the Content Security Policy Level 3 Working Draft.
+ * @constant {Set<string>}
+ * @private
+ */
+const contentSecurityPolicyDirectives = new Set([ "child-src", "connect-src", "default-src", "font-src", "frame-src", "img-src", "manifest-src", "media-src", "object-src", "prefetch-src", "script-src", "script-src-elem", "script-src-attr", "style-src", "style-src-attr", "worker-src" ]);
+
+/**
+ * The Content Security Policy fallback directives specified in the Content Security Policy Level 3 Working Draft.
+ * Property names are directive names and property values are arrays of fallback directive names.
+ * @constant {Object}
+ * @private
+ */
+const contentSecurityPolicyDirectiveFallbacks = {
+    "script-src-elem": [ "script-src-elem", "script-src", "default-src" ],
+    "script-src-attr": [ "script-src-attr", "script-src", "default-src" ],
+    "style-src-elem": [ "style-src-elem", "style-src", "default-src" ],
+    "style-src-attr": [ "style-src-attr", "style-src", "default-src" ],
+    "worker-src": [ "worker-src", "child-src", "script-src", "default-src" ],
+    "connect-src": [ "connect-src", "default-src" ],
+    "manifest-src": [ "manifest-src", "default-src" ],
+    "prefetch-src": [ "prefetch-src", "default-src" ],
+    "object-src": [ "object-src", "default-src" ],
+    "frame-src": [ "frame-src", "child-src", "default-src" ],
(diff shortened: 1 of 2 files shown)


src/userSurvey.js
```diff
@@ -38,7 +38,10 @@ import popupNoPromptPage from "./html/userSurvey.popupNoPrompt.html";
 
 permissions.check({
     module: "webScience.userSurvey",
-    requiredPermissions: [ "notifications", "webRequest" ]
+    requiredPermissions: [ "notifications", "webRequest" ],
+    requiredContentSecurityPolicy: {
+        "script-src": [ "'sha256-7MkXA5Z7wxRyznhUZN3nVVs9GEQpvRXdYihZZqR2y6w='", "'sha256-l+kgCjP15GJlVSDL9qMNffrHu8mxJcag42o2TYofOUM='" ]
```
rhelmer, line 43: Is this the hash for `userSurvey.popupNoPrompt.html`, does that mean we need to update it every time here? Seems like something we could do at build time with the rollup `replace` plugin.

src/permissions.js
```diff
@@ -5,13 +5,105 @@
...
+ *   "object-src": [ 'self' ]
+ * }
+ */
+
+/**
+ * The Content Security Policy directives specified in the Content Security Policy Level 3 Working Draft.
+ * @constant {Set<string>}
+ * @private
+ */
+const contentSecurityPolicyDirectives = new Set([ "child-src", "connect-src", "default-src", "font-src", "frame-src", "img-src", "manifest-src", "media-src", "object-src", "prefetch-src", "script-src", "script-src-elem", "script-src-attr", "style-src", "style-src-attr", "worker-src" ]);
```
rhelmer, line 23: Urgh, I mean it's fine, but it sure would be nice if these were constants we could reference from somewhere instead of having to hardcode a ton of strings (above and below too not just this set).

There is https://www.npmjs.com/package/http-headers-js but I don't know if it's really worth dragging in a third-party dependency.

rhelmer (changes requested):
This is probably all fine, wondering why we're not generating that hash at build time though. I'd like to just help double-check all these headers since we most likely have to just use strings everywhere.

src/userSurvey.js
jonathanmayer: One of the hashes is for the inline script in `userSurvey.popupNoPrompt.html`, and the other is for the inline script in `userSurvey.popupPrompt.html`.  We can change the HTML and CSS in those files without updating the hashes; only changes to the inline JS will require new hashes. While I would like us to dynamically generate the hashes eventually, I also want them to be as stable as possible... a hash change is effectively a breaking API change for library users.

src/permissions.js
jonathanmayer: I started by looking for a good quality, actively maintained, and popular library that would handle our requirements... and came up empty. That's why I implemented from scratch. 😕

jonathanmayer:
Flagging that the hashes in `userSurvey` will change when we land #53.

rhelmer (approved):

PR #70, opened 2021-04-28
Adding a pageTransition module for observing transition relationships between pages

### LINKED ISSUE #25: Add a page navigation relationship model
We have a sophisticated attention model for each page (in pageManager), but we don't yet have a good model for how users navigate between pages. We're using referrers in the news and disinformation study, which are a fine initial approximation but have significant limitations---they can be absent, or just an origin (where browsers are trending), or misleading (e.g., redirects and History API loads).

jonathanmayer:
Resolves #25.


src/webScience.js
```diff
@@ -61,3 +61,6 @@ export { randomization }
 
 import * as workers from "./workers.js"
 export { workers }
+
+import * as pageTransition from "./pageTransition.js"
+export { pageTransition }
```
rhelmer, line 66: nit, missing newline

rhelmer (changes requested):

PR #74, opened 2021-05-03
set up unit test infra

rhelmer:
I'm attempting to port the RS01 unit tests over to the web-science repo, this needs some work (it's just a direct copy right now) to make it a little more generic but I think it's a good start.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,6 +42,9 @@ jobs:
       - run:
           name: Run linting
           command: npm run lint
+      - run:
+          name: Run unit tests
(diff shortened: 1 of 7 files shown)


rhelmer:
@hamilton I'd like to get some basic infra here, what do you think about this?

I've just ported the RS01 tests over to the latest WebScience but they are otherwise mostly untouched, do you have an opinion on whether these tests are a useful starting point?

I'm also happy to land something even simpler, wdyt? I'd be happy just to have a working `npm run test` going under CircleCI with *any* test :)

PR #78, opened 2021-05-06
Implement Fathom page classification module

### LINKED ISSUE #27: Add a module for measuring DOM nodes identified by a Fathom classifier
@victwj has been implementing this functionality as part of his study.

victwj:
## Description 
Resolves #27. Integrates [Mozilla Fathom](https://mozilla.github.io/fathom/index.html) to enable ML classification of pages or DOM elements within a page. At the _element_ level, example applications include recognizing which DOM elements represent overlays, ads, login forms, privacy policies, and cookie banners. At the _page_ level, example applications include recognizing shopping pages, login pages, and news article pages. The user must first [train their own model](https://mozilla.github.io/fathom/training.html). This library then makes it simple to deploy the trained model through a Rally extension.

## Technical
The [fathom-web](https://www.npmjs.com/package/fathom-web) package is installed as devDependency. If the user wants to use this module, they must install this package in their own extension. The broad architecture of this module follows [pageManager](https://github.com/mozilla-rally/web-science/blob/main/src/pageManager.js) and [pageText](https://github.com/mozilla-rally/web-science/blob/main/src/pageText.js). A summary of how it works:

1. The `fathom.js` background script exports the `webScience.fathom.onFathomData` event. The user must call `webScience.fathom.onFathomData.addListener()` from their own background script, specifying their callback and URL match pattern. Fathom will only be run on URLs which match the specified pattern. 
2. The user must also call the global function  `window.webScience.fathom.addTrainees()` from their own content script to specify their trained Fathom model. 
3. When a tab is opened, the fathom background script sends an `webScience.fathom.isClassifiable` message to the content script running on the tab. This boolean is determined by the user's URL match patterns. 
4. Upon receiving an `isClassifiable == true` message, the Fathom content script will proceed to classify the page based on the user's rulesets, which were added via the global  `window.webScience.fathom.addTrainees()`.
5. Once the Fathom content script is done, results are sent to the background script as a `webScience.fathom.fathomData` message. The Fathom background script then processes the results and calls the user's callbacks on the results.

## Discussions

1. This module currently supports the most basic functionality: run Fathom at most once upon page load, then return the classification results. In reality, the element the user is looking for may be dynamically loaded at some point after the page is loaded. Should a rerunning function based on MutationObserver be implemented in the content script? This would be difficult to do on the user's end, so might be worth doing in this module.

2. Once an element is classified, should this module handle tracking the amount of time the element is in the user's screen? This should be doable on the user's end and is more study specific, but is likely to be a very commonly used feature, making it worth handling in the module. Most other types of element tracking (such as whether or not the element was clicked) are study specific and would fit better as a user implementation. This module can return the classified DOM elements, after which the user may monitor it as they see fit. 

## Todos

1. Memoize page contents to optimize running multiple Fathom rulesets.
2. Make border drawing a user configurable option.
3. Potentially change the use of `onUpdated` as the main event to trigger Fathom runs.
4. Write jsdoc comments.
5. Remove `console.log` lines, which are all temporary and only used for development purposes.
6. Define a proper classification results schema.

### DIFF
--- a/src/fathom.js
+++ b/src/fathom.js
@@ -27,6 +27,7 @@ import * as messaging from "./messaging.js";
 import * as events from "./events.js";
 import * as matching from "./matching.js";
 import * as inline from "./inline.js";
+import addTrainees from "./content-scripts/fathom.content.js";
 import fathomContentScript from "./content-scripts/fathom.content.js";
 
 export function test() {
@@ -107,7 +108,7 @@ let initialized = false;
  * @param {Object} options - Options for the listener.
  * @param {string[] options.matchPatterns} matchPatterns - The match patterns 
  * for pages where the listener should be notified.
- * @param {fathom.Trainees} trainees - The trainees for this listener.
+ * @param {fathom.Trainees} trainees - The trainees for this listener, a Map.
  * @private
  */
 async function addListener(listener, {matchPatterns, trainees}) {
@@ -133,7 +134,6 @@ async function addListener(listener, {matchPatterns, trainees}) {
         // Message to send to content script if this page should be classified
         messaging.registerSchema("webScience.fathom.isClassifiable", {
             isClassifiable: "boolean",
-            trainees: "object"
         })
 
         // When a tab is updated, send it a message if the page should be 
@@ -155,12 +155,13 @@ async function addListener(listener, {matchPatterns, trainees}) {
                         }
                     }
                 }
-                messaging.sendMessageToTab(tabId, {
-                    type: "webScience.fathom.isClassifiable",
-                    // isClassifiable: classifiable,
-                    isClassifiable: true,
-                    trainees: traineesSet
-                })
+                function sendIsClassifiable() {
+                    messaging.sendMessageToTab(tabId, {
+                        type: "webScience.fathom.isClassifiable",
+                        isClassifiable: classifiable,
+                    });
+                }
+                setTimeout(sendIsClassifiable, 3000); //TODO: remove
             }
         });
     }
--- a/src/fathom.js
+++ b/src/fathom.js
@@ -35,8 +35,11 @@ import fathomContentScript from "./content-scripts/fathom.content.js";
 /**
  * Fathom classification results sent by the content script.
  * @typedef {Object} FathomDataObject
- * @property {number} pageId - The ID for the page, unique across browsing sessions.
- * @property {string} url - The URL of the page, without any hash.
+ *
+ * TODO: Finalize data schema
+ * Likely would involve the URL, pageID, and the fathom classification results.
+ * Currently, just sends an object containing every element with greater than
+ * 0.5 confidence score (key), along with their feature vector and actual score (value).
  */
 
 /**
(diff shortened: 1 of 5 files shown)


src/fathom.js
```diff
@@ -0,0 +1,189 @@
...
+
+
+/**
+ * A callback function to add a fathomDataListener.
+ * @param {fathomDataCallback} listener - The listener that is being removed.
+ * @param {Object} options - Options for the listener.
+ * @param {string[] options.matchPatterns} matchPatterns - The match patterns 
+ * for pages where the listener should be notified.
+ * @private
+ */
```
jonathanmayer, line 93: Firefox recently added dynamic imports for content scripts. That would be a good way to handle rulesets and avoid the awkwardness with content script globals. Something like...
* A parameter to `addListener` specifies a path for one or more ruleset JS files and names for each ruleset.
* The ruleset JS file must be a module that exports a `ruleset()` function.
* When a page visit starts, check the URL against the match pattern set for each listener. Send an array of ruleset names and file paths to the content script.
* In the content script, import the relevant rulesets, run them, and send back the results.

src/fathom.js
```diff
@@ -0,0 +1,189 @@
...
+        
+        // Message to send to content script if this page should be classified
+        messaging.registerSchema("webScience.fathom.isClassifiable", {
+            isClassifiable: "boolean",
+        })
+
+        // When a tab is updated, send it a message if the page should be 
+        // classified with Fathom
+        // TODO: onUpdated may not be the right event
+        browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
```
jonathanmayer, line 121: Would use the `pageManager.onPageVisitStart` event so you can include a `pageId` in the message.

src/fathom.js
```diff
@@ -0,0 +1,189 @@
...
+                });
+            }
+        });
+    }
+
+    // Register content script for the listener
+    const contentScript = await browser.contentScripts.register({
+        matches: matchPatterns,
+        js: [{
+            code: inline.dataUrlToString(fathomContentScript)
```
jonathanmayer, line 144: We changed how content script imports work.

src/fathom.js
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * A callback function to handle messages from the content script.
+ * @param {FathomDataObject} fathomDataObject - Fathom classification results.
+ * @private
+ */
+function messageListener(fathomDataObject) {
+    console.log("messageListener called");
+    for (const [listener, listenerRecord] of fathomDataListeners) {
+        listener(fathomDataObject);
```
jonathanmayer, line 182: Don't want to notify the listener unless there's a match.

src/fathom.js
```diff
@@ -0,0 +1,189 @@
...
+
+        // Notify listener if the url matches the listener's matchPatterns
+        if (listenerRecord.matchPatternSet.matches(fathomDataObject.url)) {
+            listener(fathomDataObject);
+        }
```
jonathanmayer, line 187: Don't need to do this matching twice. Could associate each listener with a unique ID and include that in the message to the content script and the response.

PR #88, opened 2021-05-21
Standardization for system and monotonic clocks

jonathanmayer:
This PR updates WebScience to use one standardized clock, converting between the system clock and High Resolution Time monotonic clock as necessary.

### DIFF
--- a/src/content-scripts/pageManager.content.js
+++ b/src/content-scripts/pageManager.content.js
@@ -111,6 +111,7 @@
 
 import { generateId } from "../id.js";
 import { createEvent } from "../events.js";
+import { fromMonotonicClock } from "../timing.js";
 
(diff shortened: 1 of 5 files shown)


rhelmer (approved):
This lgtm, thanks! I know how much work went into investigating this.

The only thing I'd suggest is maybe have an API that's a little more like the new [Temporal proposal](https://tc39.es/proposal-temporal/docs/index.html), so for instance `timing.SystemClock.from(...)` and `timing.MonotonicClock.from(...)`.

This would require the timing module to be a little fancier than just exporting a handful of functions, but maybe things like `sharedMonotonicClockSupport` could be switched from a function to a getter too.

Anyway I am approving the PR as-is, you can take (or leave) the above. Thanks!

PR #102, opened 2021-06-07
Migrate from inlining to copying for asset dependencies

### LINKED ISSUE #101: Replace inlined content with copied content
We made a decision some time ago to inline WebScience content scripts, HTML, and other asset dependencies. The upside was that developers using the library wouldn't have to deal with managing WebScience files, which could be tedious and introduce both subtle versioning issues and difficult to debug dependency issues. As we've implemented the inlining approach, we've encountered a number of downsides, including requiring special Content Security Policy permissions, inability to narrowly scope CSP permissions for worker scripts, compatibility challenges for Chromium browsers, future-proofing risks for WebExtensions manifest v3, and difficulty reviewing bundled extensions. While we were starting to reconsider this decision, it became apparent that Mozilla AMO review would have hesitancy (at best) about the CSP permissions required for the inlining approach to work. So, we need to make a change.

The best (or least-bad) approach seems to be a Rollup plugin that handles WebScience file management for extension developers. This alternative approach will minimize burden and risk for developers, with the drawbacks that using WebScience would depend on using Rollup and we'd need to maintain a special Rollup plugin (since there doesn't appear to be a sufficient existing plugin).

jonathanmayer:
Resolves #101.


rhelmer (approved):
this lgtm, I appreciate the little test extension! we could pretty easily turn little one-off test extensions like this into integration tests too.

PR #118, opened 2021-07-07
[WIP, DO NOT MERGE] Add basic integration test

rhelmer:
This adds an integration test using Jest, which uses Selenium to start up Firefox and install a test extension. It also turns on extra logging in Firefox so we can see messages from the extension's background script from Selenium.

However I think it needs a few changes before we land it:

- [ ] create an extension at runtime instead of having the `test/build` dir
- [ ] make the test extension actually do something useful that can only be tested from integration vs. unit tests (maybe inject a content script as a first test?)
- [ ] add Chrome support (I think this won't be too hard so we might as well tackle it now)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,6 +43,18 @@ jobs:
           name: Run linting
           command: npm run lint
 
+  Integration testing:
+    docker:
+      - image: cimg/node:14.13.1
+    steps:
+      - checkout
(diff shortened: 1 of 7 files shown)


rhelmer:
@hamilton I think this isn't quite ready to land, but I'm pretty new to Jest and could use your opinion on this :)

Issue #238, opened 2022-04-15
detecting page unload from content script is unreliable in Chrome


rhelmer:
Per discussion in https://github.com/mozilla-rally/web-science/issues/114#issuecomment-1100225683:

> From my testing so far and discussion and discussion in [#addons on matrix](https://matrix.to/#/!CuzZVoCbeoDHsxMCVJ:mozilla.org/$ItXwR-2GCjDQIOEAn6lrYhyVh7kLWbfC7qiciqYVmwU?via=mozilla.org&via=matrix.org&via=perthchat.org), there isn't a reliable way to do this from the content script. I tested with unload which fired unreliably, also with pagehide and visibilitychange which didn't fire at all.

To make this work with Chrome in the near-term I think we need to use an API from the background to listen to this, such as listening to changes in tabs. To support manifest v3, we should do this in a way that will work with service workers so we should not assume we have access to a persistent background page.

jonathanmayer:
Possibly relevant: https://bugs.chromium.org/p/chromium/issues/detail?id=1144856

Could you say more about the testing setup? Also worth looking at the `popstate`, `freeze`, and `resume` events. Another workaround to consider is listening for these events with a script injected into the page, if the ordinary content process handles the events correctly.

rhelmer:
> Possibly relevant: https://bugs.chromium.org/p/chromium/issues/detail?id=1144856
> 
> Could you say more about the testing setup?

I'll make a minimal test case, but basically I am just putting [debugger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/debugger) statements into the content script, opening devtools on the site, and observing when it does/doesn't fire in the devtools debugger. I also have the serviceworker logging messages and it's not receiving them (note that I haven't tested this with manifest v2 just v3 under Chrome).

> Also worth looking at the `popstate`, `freeze`, and `resume` events. Another workaround to consider is listening for these events with a script injected into the page, if the ordinary content process handles the events correctly.

I'll try these, thanks. Injecting into the content process is an interesting idea, although I think getting the message back to the extension will be trickier.

rhelmer:
OK here's a branch with two commits: one where I've added a test case for manifest v2, and then a second commit which updates it to manifest v3:

https://github.com/mozilla-rally/web-science/compare/main...rhelmer:testcase-for-issue-238?expand=1

It can be used by calling `rollup -c rollup.config.js` and then run with `web-ext run -t chromium`.

[...]

PR #242, opened 2022-04-19
add integration test harness

rhelmer:
This ports over the Selenium integration tests from https://github.com/mozilla-rally/rally-web-platform and makes a few small improvements, like the ability to search the logs for a regex (which we're going to need to make https://github.com/mozilla-rally/web-science/issues/238 reproducible).

This works for Firefox as-is, but we need a few things to get Chrome totally functional:

- [ ] Chrome cannot load extensions in headless mode, so we need a display server on CI
- [ ] Selenium cannot get at Chrome log messages directly, but we can pass messages to web content

This also includes some minimal changes to the WebScience `pageManager` and `pageNavigation` modules, to make the tests work with Chrome. We'll likely find other minor incompatibilities as we add more tests.


rhelmer:
You can run the overall tests with `npm run test:integration`, but if you look at `./scripts/integration-test.sh` you can see how to change various configuration flags:

Chrome, load extension, non-headless mode
`
npm run test:integration:jest -- --test_browser=chrome --load_extension=true --headless_mode=false 2>&1 | tee integration.log 
`

Firefox, load extension, non-headless mode
`
npm run test:integration:jest -- --test_browser=firefox --load_extension=true --headless_mode=false 2>&1 | tee integration.log 
`

The `--load_extension` flag might not be that useful for WebScience, but we are probably going to want to have the ability to specify the manifest version.

src/pageManager.js
```diff
@@ -488,19 +488,18 @@ export async function initialize() {
...
             return;
         }
 
-        messaging.sendMessageToTab(tabId, {
-            type: "webScience.pageManager.pageAudioUpdate",
-            pageHasAudio: changeInfo.audible,
-            timeStamp: timing.now()
-        });
-    }, {
-        urls: [ "http://*/*", "https://*/*" ],
```
rhelmer, line 502: This filter is unsupported in Chrome, and it's not handled by the polyfill.

src/pageManager.js
```diff
@@ -521,8 +520,6 @@ export async function initialize() {
             // committed. See: https://github.com/mdn/content/issues/4469
             webNavigationTimeStamp: details.timeStamp
         });
-    }, {
```
rhelmer, line 524: This filter is supported in Chrome, but [per the docs](https://developer.chrome.com/docs/extensions/reference/webNavigation/#event-onHistoryStateUpdated):

> Conditions that the URL being navigated to must satisfy. The 'schemes' and 'ports' fields of UrlFilter are ignored for this event.

src/pageManager.js
```diff
@@ -706,14 +703,30 @@ export async function initialize() {
...
-        matches: permissions.getManifestOriginMatchPatterns(),
-        js: [{
-            file: pageManagerContentScript
-        }],
-        runAt: "document_start"
-    });
+    // Register the pageManager content script for all URLs permitted by the extension manifest.
+    // Only Firefox supports dynamic content script loading at this time.
+    try {
+        const browserInfo = await browser.runtime.getBrowserInfo();
```
rhelmer, line 709: This API appears to be [Firefox-only](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getBrowserInfo#browser_compatibility), I tried to be defensive here in case that changes in the future.

This should probably be made more abstract so it can be reused elsewhere. We know that Chrome will ship dynamic content script registration eventually, but I suspect the current API won't be polyfilled. It's more likely that the new API will be the one standardized around.

src/pageNavigation.js
```diff
@@ -147,14 +147,33 @@ async function addListener(listener, {
...
-        js: [{
-            file: pageNavigationContentScript
-        }],
-        runAt: "document_start"
-    });
+    // Only Firefox supports dynamic content script loading at this time.
+    let contentScript;
+    try {
+        // getBrowserInfo is Firefox-only as well.
+        const browserInfo = await browser.runtime.getBrowserInfo();
```
rhelmer, line 154: This is just the same check as above to try to detect the browser and handle dynamic content script registration for Firefox, or log a warning for Chrome.

tests/integration/extension/src/background.js
```diff
@@ -0,0 +1,19 @@
+/* eslint-disable no-unused-vars */
```
thomik-corp, line 1: We can probably remove this since you are using webScience below?

tsconfig.json
```diff
@@ -0,0 +1,31 @@
+{
+  "compilerOptions": {
+    "moduleResolution": "node",
+    "module": "es2020",
```
thomik-corp: We should compile to CommonJS module system. You may run into same compatibility issues with consumers otherwise as I did with SDK.

tsconfig.json
```diff
@@ -0,0 +1,31 @@
...
+    "forceConsistentCasingInFileNames": true,
+    "baseUrl": ".",
+    "allowJs": true,
+    "checkJs": true,
+    "paths": {
+      "$lib": ["src/lib"],
+      "$lib/*": ["src/lib/*"]
+    }
+  },
+  "include": ["src/**/*.d.ts", "src/**/*.js", "src/**/*.ts", "src/**/*.svelte"]
```
thomik-corp: We can probably remove most of these except .ts and .js ?

tests/integration/webarchive/index.html
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
thomik-corp: Is this file being used? I don't see it in the tests / other references.

thomik-corp: Ah I see it in follow up commit. Resolving.

src/pageManager.js
```diff
@@ -706,14 +703,30 @@ export async function initialize() {
...
+                    file: pageManagerContentScript
+                }],
+                runAt: "document_start"
+            });
+        } else {
+            console.debug("WebScience pageManager loaded, requires content script:", pageManagerContentScript);
+        }
+    } catch (ex) {
+        if (ex.message === "browser.runtime.getBrowserInfo is not a function") {
+        console.debug("WebScience pageManager loaded, requires content script:", pageManagerContentScript);
```
thomik-corp, line 723: Nit: indentation.

src/pageManager.js
```diff
@@ -706,14 +703,30 @@ export async function initialize() {
...
+        } else {
+            throw ex;
+        }
+    }
+/*
```
thomik-corp, line 728: Can probably remove this empty comment.

src/pageNavigation.js
```diff
@@ -147,14 +147,33 @@ async function addListener(listener, {
...
+            contentScript = await browser.contentScripts.register({
+                matches: matchPatterns,
+                js: [{
+                    file: pageNavigationContentScript
+                }],
+                runAt: "document_start"
+            });
+        } else {
+            console.debug("WebScience pageNavigator loaded, requires content script:", pageNavigationContentScript);
+
```
thomik-corp, line 167: Nit: can remove this extra line.

src/pageNavigation.js
```diff
@@ -147,14 +147,33 @@ async function addListener(listener, {
...
-        runAt: "document_start"
-    });
+    // Only Firefox supports dynamic content script loading at this time.
+    let contentScript;
+    try {
+        // getBrowserInfo is Firefox-only as well.
+        const browserInfo = await browser.runtime.getBrowserInfo();
+        // Register a content script for the listener
+
+        if (browserInfo.name === "Firefox") {
```
thomik-corp, line 157: Maybe later we can have a util for this:

forBrowser(Browser.Firefox, () => /... browser specific code .../, () => /... optional else branch .../);

src/pageManager.js
thomik-corp: Should we AND the condition here? audible && Firefox?

src/pageNavigation.js
rhelmer: Yes I think that would be easy to do, we had discussed doing a polyfill but it sounds like the current API isn't going to be polyfilled (maybe the new Chrome one will be implemented in Firefox instead), so I think this would be good for now.

This is going to be needed in other WS modules.

src/pageManager.js
rhelmer: Thanks, I want to run `eslint --fix` on this repo, but I didn't want to introduce noise in the review... I'll do that in a separate commit.

tsconfig.json
rhelmer: I just pushed a new tsconfig that installs and extends the `@tsconfig/recommended/tsconfig.json` package, and based it on current best practices per the typescript docs.

tsconfig.json
rhelmer: This is only for selenium right now, but that's a good point we could start using typescript for WebScience, I'll make the output commonjs.

tests/integration/extension/src/background.js
```diff
@@ -1,15 +1,52 @@
 /* eslint-disable no-unused-vars */
 import * as webScience from "@mozilla/web-science"
+import browser from "webextension-polyfill";
```
thomik-corp, line 3: What is super intriguing for me is that despite including polyfill here, I had to still include it as the first script in manifest's background section. Somewhere during bundling rollup's tree shaking is stripping things we need at the runtime in browser.

tsconfig.json
rhelmer: Oh actually the recommended config we're extending is commonjs so I think this is OK.

tests/integration/extension/src/background.js
```diff
@@ -1,15 +1,52 @@
...
+
+async function pageVisitStartListener(pageVisit) {
+    const tabs = await browser.tabs.query({
+        currentWindow: true,
+        active: true
+    });
+
+    for (let tab of tabs) {
```
thomik-corp, line 11: Minor: can use tabs.forEach

tests/integration/extension/src/background.js
rhelmer: Yes this was from the older test extension that didn't do anything except test the build (which this also does), thanks!

tests/integration/extension/src/background.js
```diff
@@ -1,15 +1,52 @@
...
 }
 
-function pageDataListener(pageData) {
+async function pageDataListener(pageData) {
+    const tabs = await browser.tabs.query({
+        currentWindow: true,
+        active: true
+    });
+
+    for (let tab of tabs) {
```
thomik-corp, line 42: We can actually refactor a method that loops and sends a message to all tabs. Then call it from here and above two places.

tests/integration/extension/src/background.js
rhelmer: Hm, I've managed to get that working... glean.js doesn't import it though and requires the polyfill at runtime :/ so I haven't been worrying about it too much. I do like having the import so I can get richer types in my IDE though.

tests/integration/extension/src/background.js
rhelmer: I'm not sure this is actually working as intended anyway. I think what we might need to do for Chrome is have a window we keep open that handles receiving these messages (this is similar to how the Firefox integration tests work).

tests/integration/ux.test.ts
```diff
@@ -0,0 +1,240 @@
...
+  if (!(arg in args)) {
+    throw new Error(`Missing required option: --${arg}`);
+  }
+}
+
+const testBrowser = args["test_browser"];
+const loadExtension = args["load_extension"] === "true";
+const headlessMode = args["headless_mode"] === "true";
+
+export let webDriver;
```
aaga, line 30: I think this should be called something like `getDriver` instead, to help differentiate from the `WebDriver` type import and the `driver` global.

Also, is this export being used somewhere?

tsconfig.json
```diff
@@ -1,31 +1,10 @@
...
-    "allowJs": true,
-    "checkJs": true,
-    "paths": {
-      "$lib": ["src/lib"],
-      "$lib/*": ["src/lib/*"]
-    }
-  },
-  "include": ["src/**/*.d.ts", "src/**/*.js", "src/**/*.ts", "src/**/*.svelte"]
-}
+  "extends": "@tsconfig/recommended/tsconfig.json",
```
thomik-corp, line 2: This is awesome! I confirmed this uses CommonJS !

tests/integration/ux.test.ts
```diff
@@ -0,0 +1,240 @@
...
+
+// Wait ten minutes overall before Jest times the test out.
+jest.setTimeout(60 * 10000);
+
+let driver: WebDriver;
+let screenshotCount = 0;
+let server: http.Server;
+
+const PORT = 8000;
+const URL = `http://localhost:${PORT}`;
```
aaga, line 54: rename `BASE_URL`

thomik-corp (approved):
Left some comments but code looks good overall. Approving this and you can address those changes and merge!

tests/integration/extension/src/background.js
rhelmer: Yes I agree, I removed this for now since Chrome support needs a little more work and I want to split it out to a separate PR, and focus on getting the infrastructure. Getting log messages out of extensions is a real hassle, I want to research it a little more.

src/pageNavigation.js
rhelmer: Hm I don't mind doing this now, but I don't see a great place right now in this code for shared utilities :)

tests/integration/ux.test.ts
```diff
@@ -0,0 +1,240 @@
...
+        testBrowser,
+        message
+      ),
+    WAIT_FOR_PROPERTY
+  );
+}
+
+describe("WebScience Test Extension", function () {
+  beforeAll(async () => {
+    server = http.createServer((req, res) => {
```
aaga, line 71: Not sure if we are planning to add more complex pages to the webarchive, but if so I think it makes sense to switch to something like the below to allow for folder structure.
```
// npm i -D http-server
import { spawn } from "child_process";
let http-server;
...
beforeAll(() => {
    http-server = spawn("http-server", PATH, "-p", PORT);
}

afterAll(() => {
    http-server.kill();
}

aaga (approved):
mostly style/future-proofing comments. I think it's ready to land tho!

tsconfig.json
aaga: 🎉

src/pageNavigation.js
rhelmer: OK I've created a `utils.js` and made a `registerContentScripts()` function, which should be easier to reuse.

tests/integration/ux.test.ts
rhelmer: Yeah, this is used below to choose the right underlying driver (Firefox or Chrome) based on command-line flags. It would be nice to move this to utils or something and make it easier to share across repos.

tests/integration/ux.test.ts
rhelmer: Yes I think we will, one thing we've been working on elsewhere is making it possible for folks to build their own local archive for test purposes: https://github.com/mozilla-rally/facebook-pixel-hunt/blob/main/tests/integration/webarchive/create-archive.sh

I think using a webserver that supports static files is a good idea, I was trying to avoid dragging in dependencies for thhis first update but I think it's worth it.

src/pageManager.js
rhelmer: Well we can still do it, but we need to handle it inside the callback vs. setting a filter on the `addListener` call. I just cleaned this up a little and corrected the way it works, it should be cross-browser now.

rhelmer:
Got an OK from @jonathanmayer, renamed `utils.js` to `contentScripts.js` and will file a followup to make a rollup plugin to generate output for `manifest.json`.

rhelmer:
The current state is that we run integration tests for `pageManager` and `pageNavigation` under Firefox in CI, and Chrome can be run locally (cannot be run headless, Chrome extensions do not work in this mode).

Filed https://github.com/mozilla-rally/web-science/issues/243 and https://github.com/mozilla-rally/web-science/issues/244 as followups.

PR #247, opened 2022-04-20
run Chrome non-headless with xvfb-run in CI

rhelmer:
I figured out a very easy way to run this with Xvfb as a display server! It's also possible to attach a VNC server via an ssh tunnel for debugging, but Selenium also takes screenshots (and it can be run locally) so we shouldn't usually need to.

### DIFF
--- a/scripts/integration-test.sh
+++ b/scripts/integration-test.sh
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -eo pipefail
 
+XVFB_RUN() {
+    if hash xvfb-run 2>/dev/null
+    then
+        xvfb-run "$@"
+    else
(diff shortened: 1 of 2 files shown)


scripts/integration-test.sh
```diff
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -eo pipefail
 
+XVFB_RUN() {
+    if hash xvfb-run 2>/dev/null
+    then
+        xvfb-run "$@"
+    else
+        echo "WARN: xvfb-run must be installed to run Chrome headless"
```
aaga, line 9: nit: update to "run Chrome with extensions enabled on a headless server" or something to disambiguate headless driver from headless operation

aaga (approved):

PR #248, opened 2022-04-20
Check for all integration logs

rhelmer:
The logic in the `extensionLogsPresent` function isn't quite what we want after I looked into it further - previously it just looked for a single regex to match at least one line in the logs, but now we want to pass a list of regexes to match, and only return true if *all* have a match.

### DIFF
--- a/tests/integration/utils.ts
+++ b/tests/integration/utils.ts
@@ -82,7 +82,7 @@ export async function extensionLogsPresent(
         }
       }
       if (!found) {
-        return false;
+        throw new Error(`Failed to find match: ${match} in Chrome logs`);
       }
     }
   } else if (testBrowser === "firefox") {
@@ -95,7 +95,7 @@ export async function extensionLogsPresent(
         found = true;
       }
       if (!found) {
-        return false;
+        throw new Error(`Failed to find match: ${match} in integration.log`);
       }
     }
   } else {
--- a/tests/integration/utils.ts
+++ b/tests/integration/utils.ts
@@ -61,8 +61,8 @@ export async function findAndAct(
  *        WebDriver in use.
  * @param {string} testBrowser
  *        Browser in use.
- * @param {RegExp} message
- *        Message to search for.
+ * @param {Array<RegExp>} message
+ *        Messages to search for. All messages must be present.
  * @returns {Promise<boolean>}
  *        Whether or not the message was found.
  */
--- a/tests/integration/utils.ts
+++ b/tests/integration/utils.ts
@@ -87,6 +87,7 @@ export async function extensionLogsPresent(
     }
   } else if (testBrowser === "firefox") {
     const fileBuffer = await fs.promises.readFile("./integration.log");
+
     // FIXME it would be more efficient to keep track of where we are in the log vs. re-reading it each time.
     // FIXME this would also make it more like the behavior of Chrome's log interface.
     for (const match of matches) {
--- a/tests/integration/utils.ts
+++ b/tests/integration/utils.ts
@@ -91,11 +91,7 @@ export async function extensionLogsPresent(
     // FIXME it would be more efficient to keep track of where we are in the log vs. re-reading it each time.
     // FIXME this would also make it more like the behavior of Chrome's log interface.
     for (const match of matches) {
-      let found = false;
-      if (match.test(fileBuffer.toString())) {
-        found = true;
-      }
-      if (!found) {
+      if (!match.test(fileBuffer.toString())) {
         throw new Error(`Failed to find match: ${match} in integration.log`);
       }
     }
--- a/tests/integration/utils.ts
+++ b/tests/integration/utils.ts
@@ -73,29 +73,36 @@ export async function extensionLogsPresent(
 ): Promise<boolean> {
   if (testBrowser === "chrome") {
     const logEntries = await driver.manage().logs().get(logging.Type.BROWSER);
-    let found = false;
-    for (const logEntry of logEntries) {
-      for (const match of matches) {
+
+    for (const match of matches) {
+      let found = false;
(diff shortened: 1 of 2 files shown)


tests/integration/utils.ts
```diff
@@ -73,29 +73,36 @@ export async function extensionLogsPresent(
 ): Promise<boolean> {
```
aaga, line 73: just noticed, the function param description above (line 64) should be updated for the new array of RegExp matches instead of a single message

tests/integration/utils.ts
```diff
@@ -73,29 +73,36 @@ export async function extensionLogsPresent(
...
+      }
     }
-    return found;
   } else if (testBrowser === "firefox") {
     const fileBuffer = await fs.promises.readFile("./integration.log");
-    let found = false;
     // FIXME it would be more efficient to keep track of where we are in the log vs. re-reading it each time.
     // FIXME this would also make it more like the behavior of Chrome's log interface.
     for (const match of matches) {
+      let found = false;
```
aaga, line 93: I know things might change if the FIXME above gets implemented, but for now the logic can be simplified to the below and I think is easier to follow:
```
for (const match of matches) {
  if (!match.test(fileBuffer.toString())) {
    return false;
  }
}
```

aaga:
(minor suggestions, but I agree with logic change)

tests/integration/utils.ts
aaga: @rhelmer are you planning to change this one? if not I can go ahead and approve as is

rhelmer: @aaga oh I did it this way so the non-matching pattern can be printed below

aaga: hmm I don't see where the printing is happening. But also the change I'm suggesting doesn't change any aspect of control flow, just gets rid of the extraneous `found` bool (it's not needed in this case since unlike the chrome case there is no nested for loop)

rhelmer: Sorry, I meant the exception message.

rhelmer: I changed this so it throws an exception with a message about the first message that failed to match vs. returning `false`, this is to make the test bail out earlier and show a more meaningful stack trace.

rhelmer: oh now that i think about it though i could still take your suggestion, just throwing vs. returning

aaga: @rhelmer oh yes I understand the change you made to return false as soon as a match fails. The change I was suggesting above keeps that same new behavior, but without the need for the extra variable.

rhelmer: Oh hm actually I don't think changing this will work, we need to check over all log entries for matches, and only error out if there were no matches... otherwise this would fail on the first log entry that didn't match, right?

aaga: from my understanding, in the chrome case we need to loop through all log entries before throwing. but for firefox, all the logs are already bunched together in one file, so we don't need to wait, we can just throw as soon as any match isn't found in the mega-file:

```
for (const match of matches) {
  if (!match.test(fileBuffer.toString())) {
    throw new Error(`Failed to find match: ${match} in integration.log`);
  }
}
```

aaga: @rhelmer  actually thinking about it more, I don't know if we can throw here. this `extensionLogsPresent` gets called inside an `await driver.wait(...)`, which poles every 0.5 seconds for a truthy value. I don't think it's designed to catch a throw, only to retry on a falsy value being returned

rhelmer: > from my understanding, in the chrome case we need to loop through all log entries before throwing. but for firefox, all the logs are already bunched together in one file, so we don't need to wait, we can just throw as soon as any match isn't found in the mega-file:
> 
> ```
> for (const match of matches) {
>   if (!match.test(fileBuffer.toString())) {
>     throw new Error(`Failed to find match: ${match} in integration.log`);
>   }
> }
> ```

Oh right I see what you mean, and yes that's correct.



> @rhelmer actually thinking about it more, I don't know if we can throw here. this `extensionLogsPresent` gets called inside an `await driver.wait(...)`, which poles every 0.5 seconds for a truthy value. I don't think it's designed to catch a throw, only to retry on a falsy value being returned

Hm, pretty sure I tried and it just halted the test entirely which was what I wanted in this case. I'll double-check though.

rhelmer: I just tested and it throws and aborts the test (since nothing tries to handle it), which I think makes it more obvious what the failure was:

```
FAIL tests/integration/ux.test.ts (11.809 s)
  WebScience Test Extension
    ✕ tests navigation by modifying address bar in a single tab (2950 ms)
    ✓ tests navigation by loading each site in a new window (4237 ms)
    ✓ tests navigation by loading each site in a new tab (3735 ms)

  ● WebScience Test Extension › tests navigation by modifying address bar in a single tab

    Failed to find match: /(WebScienceTest - test123 Page visit stop).*(http:\/\/localhost:8000)/ in integration.log

      93 |     for (const match of matches) {
      94 |       if (!match.test(fileBuffer.toString())) {
    > 95 |         throw new Error(`Failed to find match: ${match} in integration.log`);
         |               ^
      96 |       }
      97 |     }
      98 |   } else {

      at extensionLogsPresent (tests/integration/utils.ts:95:15)
      at tests/integration/ux.test.ts:73:7

Test Suites: 1 failed, 1 total
Tests:       1 failed, 2 passed, 3 total
Snapshots:   0 total
Time:        11.879 s, estimated 12 s
Ran all test suites matching /.\/tests\/integration\//i.
```

aaga:
waiting on incorporating Firefox logs update before merging, as per discussion with @rhelmer

PR #250, opened 2022-04-21
Fix issue #238, use pagehide not unload listener

rhelmer:
This seems to work fine in Firefox and (mostly) in Chrome, but it intermittently fails to fire in Chrome.

Per MDN, the [visibilitychange event is preferred](https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event#usage_notes), but it does note the back/forward cache behavior. Using `pagehide` when that isn't available is suggested as well.

I think the root cause might be that the back/forward cache doesn't get canceled in Chrome the way it does in Firefox, but I'm having a lot of trouble reproducing it in a test environment. Using `pagehide` instead seems to work when I test sites in the wild, and it still passes the test suite.

### DIFF
--- a/src/content-scripts/pageManager.content.js
+++ b/src/content-scripts/pageManager.content.js
@@ -422,8 +422,7 @@ import { fromMonotonicClock } from "../timing.js";
     // Send the page visit stop event on the window unload event,
     // using the timestamp for the unload event on the global
     // monotonic clock 
-    window.addEventListener("unload", (event) => {
+    window.addEventListener("pagehide", (event) => {
         pageVisitStop(fromMonotonicClock(event.timeStamp, true));
     });
-    
 })();


rhelmer:
Hm, on further testing I don't Chrome supports `document.visibilityState === 'unloaded'` and `document.visibilityState === 'hidden'` isn't quite what we want, but I'll see what I can determine.

aaga (approved):
confirmed test passes locally for me as well.

@rhelmer just want to check my understanding from reading the linked issue: the plan is to eventually use tabs.onChange, but we're doing it this way for now for MVP?

rhelmer:
> confirmed test passes locally for me as well.
> 
> @rhelmer just want to check my understanding from reading the linked issue: the plan is to eventually use tabs.onChange, but we're doing it this way for now for MVP?

Yes, I think ultimately depending on the content script environment is going to mean depending on implementation details and we won't get predictable cross-browser behavior there anytime soon. The `tabs.*` APIs on the other hand are more amenable to standardizing.

PR #252, opened 2022-04-22
only call unregister for valid dynamic content scripts

rhelmer:


src/contentScripts.js
```diff
@@ -4,6 +4,8 @@
  *
  * @param {Array<String>} matchPattern
  * @param {import} contentScript
+ *
+ * @returns {(browser.contentScripts.RegisteredContentScript|null)} - a registered content script if Firefox, null if otherwise.
```
aaga, line 8: I think currently this functions returns undefined, not null. Probably makes sense to return null explicitly in the else statement?

aaga (changes requested):

src/contentScripts.js
rhelmer: Good catch thanks :) Can't wait to start using ts here

aaga: amen

aaga (approved):

PR #278, opened 2022-06-22
export contentScripts

rhelmer:
I'd like to be able to use this over in https://github.com/mozilla-rally/study-template/pull/267, as a cross-browser way to register content scripts.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
                 "@types/jest": "^27.4.1",
                 "@types/minimist": "^1.2.2",
                 "@types/selenium-webdriver": "^4.0.18",
-                "chromedriver": "^101.0.0",
+                "chromedriver": "^103.0.0",
                 "eslint": "^8.1.0",
                 "eslint-plugin-import": "^2.22.1",
                 "eslint-plugin-mocha": "^10.0.1",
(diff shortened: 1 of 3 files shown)


aaga (approved):
ignore; see below

aaga (changes requested):
Looks like there is some Chrome version error in CircleCI?

You mentioned that you want to export this for cross-browser content script things, but it looks like this only works for Firefox, which I think you know! So just for my own understanding, are we doing this for posterity i.e. for when Chrome eventually supports it? or are we going to use it right off the bat with Firefox?

rhelmer:
> Looks like there is some Chrome version error in CircleCI?

Thanks, looks like it's because there's a new Chrome version out and I need to bump the ChromeDriver selenium package: 

```
SessionNotCreatedError: session not created: This version of ChromeDriver only supports Chrome version 101
    Current browser version is 103.0.5060.53 with binary path /usr/bin/google-chrome
```

> Also, you mentioned that you want to export this for cross-browser content script things, but it looks like this only works for Firefox, which I think you know! So just for my own understanding, are we doing this for posterity i.e. for when Chrome eventually supports it? or are we going to use it right off the bat with Firefox?

So my thought here is that we'll use it now which will work in Firefox, and it'll cause a message to be logged for Chrome (just a notification that a content script needs to be loaded). The next step would be to hook up a rollup JSON plugin (https://github.com/mozilla-rally/web-science/issues/281) that generates the needed manifest.json entries for Chrome, instead of simply logging.

Eventually there will be a new cross-browser scripting API released by Google, we can switch away from needing this wrapper once both browsers support it, but we may end up needing it for the transition period (if for instance Firefox is still using the old API and Chrome has the new for a while).

rhelmer:
I'm going to go ahead and land this, since it's not really much of a change (just exporting existing functions) now that tests pass, since it's blocking getting the study-template changes landed.

Issue #284, opened 2022-06-30
Provide method for content scripts to only run on certain domains in Chrome


rhelmer:
Firefox supports [dynamically loading content scripts](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/contentScripts/register), but until Chrome ships [an equivalent API](https://bugs.chromium.org/p/chromium/issues/detail?id=1054624) we don't have a way to restrict content scripts to only run on certain domains (specified as match patterns).

In PR https://github.com/mozilla-rally/web-science/issues/281 we attempted to write a roll-up plugin to generate entries , but [Chrome cannot handle the ~4k sample of domains](https://github.com/mozilla-rally/web-science/issues/281#issuecomment-1168257836) from the Beyond the Paywall study that I tested.

A few approaches come to mind:

1. content scripts could message-pass with the background script/serviceworker to determine if they should run
2. a list of allowed domains could be inlined into the content scripts

Since WebScience is built by the caller, I think both of these are pretty easy to do. The trade-offs are that (1) involves more messaging and runtime overhead, and (2) will cause content scripts to be larger on average, but no messaging would be required and the content scripts could run totally standalone.

(1) is probably the most flexible, and it might end up mattering if there are several content scripts that filter on a very large set of match patterns.

rhelmer:
It looks like this might be supported in Chrome now, Manifest v3 only:

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/registerContentScripts#browser_compatibility

Firefox supports it in MV3 as of 101, and MV2 as of 102+, so I think we might be able to switch to this now.

PR #301, opened 2022-07-16
fixes issue #284, use new cross-browser scripting api for registering…

rhelmer:
… content scripts

Review expectations: this has integration tests for Firefox (mv2) and Chrome (mv3), so I think just a visual check is OK. If you'd like to try loading this into an extension or load the test extension into your browser that's fine too (but probably redundant with the integration test).


src/pageManager.js
```diff
@@ -704,7 +703,15 @@ export async function initialize() {
...
     // Register the pageManager content script for all URLs permitted by the extension manifest.
-    contentScripts.registerContentScript(permissions.getManifestOriginMatchPatterns(), pageManagerContentScript);
+    const matchPatterns = permissions.getManifestOriginMatchPatterns();
+
+    // TODO register this content script so it may be unloaded later. 
+    await browser.scripting.registerContentScripts([{
+        id: "pageManager",
+        js: ["dist/browser-polyfill.min.js", pageManagerContentScript],
+        matches: matchPatterns,
+        persistAcrossSessions: false
```
aaga, line 713: the helper this is replacing had `runAt: document_start` set. It seems that [the default](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extensionTypes/RunAt) is `document_idle`. Should we keep `document_start`? I wonder if it might make a difference in catching certain events etc?

src/pageNavigation.js
```diff
@@ -148,7 +147,12 @@ async function addListener(listener, {
 
     // Compile the match patterns for the listener
     const matchPatternSet = matching.createMatchPatternSet(matchPatterns);
-    const contentScript = await contentScripts.registerContentScript(matchPatterns, pageNavigationContentScript);
+    const contentScript = await browser.scripting.registerContentScripts([{
```
aaga, line 150: nit: on line 38 of this file, the @property type should now be changed from `browser.contentScripts.RegisteredContentScript` to `browser.scripting.RegisteredContentScript`

aaga, line 150: Actually, on that note, this new function returns an array instead of a single object. Should this be `const [contentScript] = ...` instead?

aaga: We are storing it in the `pageDataListeners` below. I'm not sure if that's used anywhere, but seems wise to be consistent for posterity

src/permissions.js
```diff
@@ -214,5 +214,21 @@ export function getManifestOriginMatchPatterns() {
             }
         }
     }
+
+    // Manifest v3 requires host match patterns to be in an array under the host_permissions key.
+    if ("manifest_version" in manifest && manifest.manifest_version === 3) {
```
aaga, line 219: Any reason not to if/else this with the above? or alternatively, combine with something like this:

```
const manifestPermissions = [ ];

("permissions" in manifest) &&
Array.isArray(manifest.permissions) &&
manifestPermissions.push(...manifest.permissions);

("host_permissions" in manifest) &&
Array.isArray(manifest.host_permissions) &&
manifestPermissions.push(...manifest.host_permissions);

for (const permission of manifestPermissions) {
...
```

Note, I also don't think it's necessary to check for the manifest version? But I understand wanting to have some indication for clarity (maybe in a comment).

src/pageNavigation.js
```diff
@@ -148,7 +147,12 @@ async function addListener(listener, {
...
+        id: "pageNavigation",
+        js: ["dist/browser-polyfill.min.js", pageNavigationContentScript],
+        matches: matchPatterns,
+        persistAcrossSessions: false
+    }]);
```
aaga, line 155: also same question as above re: `runAt: document_start`

tests/integration/extension/src/background.js
```diff
@@ -25,4 +28,14 @@ async function pageDataListener(pageData) {
...
\ No newline at end of file
+webScience.pageNavigation.onPageData.addListener(pageDataListener, { matchPatterns: ["<all_urls>"] });
+
+// Load content script(s) required by this extension.
+browser.scripting.registerContentScripts([{
+    id: "webextension-test",
+    js: ["dist/browser-polyfill.min.js", "dist/test.content.js"],
+    matches: ["<all_urls>"],
+    persistAcrossSessions: false
+}])
```
aaga, line 39: same question as above re: `runAt: document_start`
(see the deleted line 31 from original test `manifest.json`)

aaga (changes requested):
The meat of it looks good! A couple functionality questions, and a couple small style things.

aaga:
Another thing we should address: it might be worth it to add a "Chrome/MV3 support" column to that modules table in the README? since it seems we'll be updating these modules piecemeal as and when we need them. Doesn't necessarily need to be in the scope of this PR, but just wanted to put it on the radar as something that would be useful to consumers of this library (primarily future us!).

src/pageManager.js
aaga: see: https://github.com/mozilla-rally/web-science/blob/043176f033642f6e3d708fab2311aa6b6e3b036a/src/contentScripts.js#L22****

src/pageNavigation.js
aaga: see: https://github.com/mozilla-rally/web-science/blob/75f8515b45c75d3d7cfda651cd7b6e7424e1600e/src/pageNavigation.js#L38

tests/integration/extension/src/background.js
aaga: see: https://github.com/mozilla-rally/web-science/blob/043176f033642f6e3d708fab2311aa6b6e3b036a/tests/integration/extension/manifest.json#L31

src/pageManager.js
rhelmer: Ugh, sorry so apparently I fixed this in https://github.com/mozilla-rally/web-science/pull/302 which is based on this PR, I just cherry-picked that commit over.

src/pageNavigation.js
rhelmer: Yeah, I don't think this is really working as intended, but I'll go ahead and change it anyway :) I'd like to make it possible to unload all of these content scripts but it doesn't look like everything is wired up correctly right now.

src/permissions.js
rhelmer: So the reason we check for manifest version is that in v2, match patterns went in `permissions` like:

```
permissions: [
  "tabs",
  "...",
  "<all_urls>"
]
```

Where in manifest v3, there's a new `host_permissions` to hold just match patterns.

So I think what we want is in mv2 to continue to look in `permissions`, and in mv3 use `host_permissions` instead.

I'll add a comment, does that all sound OK and correct @aaga ?

rhelmer: This is covered in https://developer.chrome.com/docs/extensions/mv3/intro/mv3-migration/#host-permissions among other docs

rhelmer:
> Another thing we should address: it might be worth it to add a "Chrome/MV3 support" column to that modules table in the README? since it seems we'll be updating these modules piecemeal as and when we need them. Doesn't necessarily need to be in the scope of this PR, but just wanted to put it on the radar as something that would be useful to consumers of this library (primarily future us!).

Good idea, did this as well... I made it a single "Browser Support" column with the text "Firefox" and/or "Chrome", we could use browser icons later if we add more and need to save space, but this seemed simpler and more readable.

src/permissions.js
aaga: oh yes I understood that part, sorry I should have been more verbose with my comment.

my suggestion is mostly about structure not functionality. currently we have two identical for loops, just looping over different arrays. It makes sense to me to combine those two arrays and just have one loop. Additionally, for the manifest v3 case, the code still first loops over the non-host permissions, even though we know there will be nothing of interest there.

tests/integration/extension/src/background.js
```diff
@@ -25,4 +28,15 @@ async function pageDataListener(pageData) {
 webScience.pageManager.onPageVisitStart.addListener(pageVisitStartListener);
 webScience.pageManager.onPageVisitStop.addListener(pageVisitStopListener);
 
-webScience.pageNavigation.onPageData.addListener(pageDataListener, { matchPatterns: ["<all_urls>"] });
\ No newline at end of file
+webScience.pageNavigation.onPageData.addListener(pageDataListener, { matchPatterns: ["<all_urls>"] });
+
```
aaga, line 32: side note unrelated to this PR: do we need to worry about these `addListener` and `registerContentScripts` functions being called multiple times due to restarts of the service worker in mv3?

i.e. with mv3 we need to start ensuring our background scripts are "idempotent", right?

src/linkExposure.js
```diff
@@ -72,7 +72,7 @@ const ignoreSelfLinks = true;
  * @property {matching.MatchPatternSet} linkMatchPatternSet - The match patterns for link URLs.
  * @property {matching.MatchPatternSet} pageMatchPatternSet - The match patterns for pages.
  * @property {boolean} privateWindows - Whether to report exposures in private windows.
- * @property {browser.contentScripts.RegisteredContentScript} contentScript - The content
+ * @property {browser.scripting.RegisteredContentScript} contentScript - The content
```
aaga, line 75: hmm technically this one should stay as it was, since we're still using the old `browser.contentScripts` API in this file

aaga: I looked into it and they're actually (confusingly) very different objects.
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/RegisteredContentScript
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/contentScripts/RegisteredContentScript

the `contentScripts` one has the `unregister` method built into it, and no other info. the `scripting` seems to be the opposite: lots of info about the script but no built-in method to unregister (you have to call `scripting.unregisterContentScripts` instead)

src/pageTransition.js
```diff
@@ -193,7 +193,7 @@ import pageTransitionClickContentScript from "include:./content-scripts/pageTran
  * @property {boolean} privateWindows - Whether to notify the listener about page transitions in
  * private windows and whether to consider pages loaded in private windows when generating
  * time-based transition information.
- * @property {browser.contentScripts.RegisteredContentScript} contentScript - The content
+ * @property {browser.scripting.RegisteredContentScript} contentScript - The content
```
aaga, line 196: same as `linkExposure` above, this should remain `contentScripts`

src/permissions.js
rhelmer: Oh, I see, thanks.

tests/integration/extension/src/background.js
rhelmer: That's a good question, I'll look into this. It's also an issue with updates, but service worker restarts are much more frequent obviously :)

src/linkExposure.js
rhelmer: Yeah I'd like to file a followup on this, I don't think WebScience is unregistering these reliably as-is (since I found at least once place it wasn't registering the script).

src/pageTransition.js
rhelmer: Oh thanks yeah blind search/replace on my part, let's leave these alone until they are ported.

src/permissions.js
rhelmer: OK I updated it as you suggested, although it occurs to me that in MV3 there should never be any host match patterns in the permissions so it's a bit of wasted effort, but this list is likely to be very small in every imaginable case so I think over-optimizing isn't worth it vs. code clarity (n will always be small).

Also the MV2 bits of this code can go away when Firefox transitions.

src/linkExposure.js
rhelmer: I was pretty confused about this too, I think `scripting.unregisterContentScripts` is the new way to do this. `scripting.getRegisteredContentScripts` can be used to get the list of content scripts so I think all we really need to keep around is the ID that we register the scripts with, we don't need to keep the reference around.

rhelmer: Filed https://github.com/mozilla-rally/web-science/issues/309 to follow up on this

aaga (approved):
let's gamble, try merging! jk

tests/integration/extension/src/background.js
rhelmer: So `registerContentScripts` will fail if we attempt to re-use an ID, so I think those are fine.

I think in the extension update case I brought up the browser will tear down existing listeners, so that's probably not really an issue. They are infrequent enough (and if it is a problem, it's a pre-existing problem) so I'm going to set that aside but it's worth looking into.

I don't think the background script is re-initialized every time it's woken up, it works more like this:

1. browser evaluates and runs the service worker on extension startup
2. service worker adds a `stateChangeCallback` listener in rally SDK
3. service worker initializes Glean, and the script finishes

At some point, the `stateChangeCallback` is called which puts the extension into the `run` state, and it registers some additional listeners. This is where the `WebScience` listeners and any others are added.

The "waking up" that the service worker does is to do things like run callbacks from these already-registered listeners; I don't believe that it does steps 1-3 again until the extension is restarted. Generally an extension restart happens on browser restart or on extension update (which, unless we explicitly delay it with [runtime.onUpdateAvailable](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/onUpdateAvailable), will happen immediately and restartlessly). `chrome://extension` developer mode and `about:debugging` in Firefox can force a restart too.

aaga: ok yes this is making some sense to me. Although for now I'm still trying to wrap my head around the non-Rally case, since we're looking at the WebScience test extension.

I'm realizing I maybe have a fundamental misunderstanding about how service workers work. The top-level code, that's only run once on extension start up? And then subsequent "wakes" only happen if the top-level code registered listeners, and the listener gets triggered? In which case, I think my concern is moot because the top-level code is not run again when the service worker awakes.

BUT what confused me is language in, for example, [this docs section](https://developer.chrome.com/docs/extensions/mv3/migrating_to_service_workers/#event_listeners) for Chrome, which says "When a service worker is terminated, so are the event listeners associated with it." Which conflicts with my understanding above.

I've approved the PR, but I do still want to understand this better for future.

rhelmer: OK I tested it and I think you're right @aaga - it does look like the service worker gets re-evaluated. As long as the listeners are torn down first I don't *think* it should matter (and content scripts shouldn't get re-registered due to the ID). Nice find, thanks!

Issue #280, opened 2022-06-23
pageText (article detection and content capture) not working on Chrome


rhelmer:
`pageText` tries to pass filters to `browser.tabs.onUpdated.addListener` which is Firefox-only, filtering needs to be done in the callback instead.

rhelmer:
OK, so Firefox has an `isArticle` filter that this is using specifically:

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/onUpdated#isarticle

>isArticle Optional
>    boolean. True if the tab is an article and is therefore eligible for display in [Reader Mode](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/toggleReaderMode).

I think we'll need to instead handle this in the callback, and use the Readability library (which is a standalone version of the code that powers Firefox's reader mode, and we already ship it in beyond-the-paywall. There's [an npm package](https://www.npmjs.com/package/@mozilla/readability))

rhelmer:
> OK, so Firefox has an `isArticle` filter that this is using specifically:
> 
> https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/onUpdated#isarticle
> 
> > isArticle Optional
> > boolean. True if the tab is an article and is therefore eligible for display in [Reader Mode](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/tabs/toggleReaderMode).
> 
> I think we'll need to instead handle this in the callback, and use the Readability library (which is a standalone version of the code that powers Firefox's reader mode, and we already ship it in beyond-the-paywall. There's [an npm package](https://www.npmjs.com/package/@mozilla/readability))

Specifically the `isProbablyReaderable()` function.

rhelmer:
> 

Small correction - we can't directly get the page contents from the listener (which is in the background script/worker), so instead I put the `probablyIsReaderable()` call in the content script, which does have direct access to the DOM of the web page. It does an early-return and does not send a message if this call is `false`.

PR #302, opened 2022-07-29
Cross platform article detection

rhelmer:
Fixes issue #280.

This depends on https://github.com/mozilla-rally/web-science/pull/301, which should land first.

Review expectations: this is covered by integration tests, so just visual inspection is probably fine. Caveat is that the tests just ensure that the message is making it all the way back from the content script -> background script, the actual message contents aren't thoroughly checked (yet, I think we should start parsing these messages instead of using regexes from test code, they are stringified JSON).

One bit of logic I changed here is that Firefox supports an `isArticle` filter on `browser.tabs.onUpdated.addListener`, but Chrome neither supports filters (aka `extraParams`) nor does it support `isArticle` (which tries to guess if the content is likely to be parseable by Reader Mode in Firefox).

So now, a message is *always* passed to the `pageText` content script when the `pageText` background module is loaded, and the content script instead decides if the text seems likely to work in readability (using the `@mozilla/readability` library which is a standalone library containing the code used by Firefox Reader Mode). The content script will exit early and not send a message back if that's the case.

All that said, I think this is probably good enough for starters, but I'll file a few followup issues to make the testing more robust:

- tests should actually analyze the result and ensure the actual test page contents match expected
- we should have a few test pages that fail the readability `isProbablyReaderable()` test, to ensure that this check is working and the content script aborts as early as possible

### DIFF
--- a/src/pageText.js
+++ b/src/pageText.js
@@ -54,7 +54,6 @@ import * as messaging from "./messaging.js";
 import * as matching from "./matching.js";
 import * as events from "./events.js";
 import * as pageManager from "./pageManager.js";
-import * as permissions from "./permissions.js";
 import pageTextContentScript from "include:./content-scripts/pageText.content.js";
 
 /**
--- a/src/pageText.js
+++ b/src/pageText.js
@@ -170,25 +170,25 @@ async function addListener(listener, {
                 }
             }
         },
-        {
-            type: "webScience.pageText.parsedText",
-            schema: {
-                pageId: "string",
-                url: "string",
-                title: "string",
-                content: "string",
-                textContent: "string",
-                privateWindow: "boolean"
-            }
-        });
+            {
+                type: "webScience.pageText.parsedText",
+                schema: {
+                    pageId: "string",
+                    url: "string",
+                    title: "string",
+                    content: "string",
+                    textContent: "string",
+                    privateWindow: "boolean"
+                }
+            });
 
         // Notify the content script when there is a new Readability status
         // for a page and the page URL matches at least one listener
         messaging.registerSchema("webScience.pageText.isArticle", {
             isArticle: "boolean"
         });
-        browser.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
-            if("isArticle" in changeInfo && "url" in tab) {
+        browser.tabs.onUpdated.addListener((tabId, _changeInfo, tab) => {
+            if ("url" in tab) {
                 // Test match patterns here rather than in the tabs.onUpdated
                 // listener options so we don't have to manage multiple listeners
                 // or remove and add the listener while events might be queued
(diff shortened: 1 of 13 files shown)


src/pageText.js
```diff
@@ -170,54 +169,57 @@ async function addListener(listener, {
...
-            schema: {
-                pageId: "string",
-                url: "string",
-                title: "string",
-                content: "string",
-                textContent: "string",
-                privateWindow: "boolean"
-            }
-        });
+            {
```
aaga, line 172: did you mean to add this indentation here?

src/pageText.js
```diff
@@ -170,54 +169,57 @@ async function addListener(listener, {
...
+            if ("url" in tab) {
                 // Test match patterns here rather than in the tabs.onUpdated
                 // listener options so we don't have to manage multiple listeners
                 // or remove and add the listener while events might be queued
                 for (const listenerRecord of textParsedListeners.values()) {
                     if (listenerRecord.matchPatternSet.matches(tab.url)) {
                         messaging.sendMessageToTab(tabId, {
                             type: "webScience.pageText.isArticle",
-                            isArticle: tab.isArticle
+                            isArticle: true
```
aaga, line 198: hmm this feels confusing. I guess I'm not really sure why it was like this in the first place. really the schema should remove this boolean since we're not checking for article-ness here anymore, right? and the explanation comments in `pageText.content.js` should be updated to reflect the new logic.

I just don't want our (currently good) code legibility to indicate one thing, but then actually things work differently than you expect

aaga (changes requested):

src/pageText.js
aaga: These are the comments I'm referring to:

https://github.com/mozilla-rally/web-science/blob/8fe8c055b6c87f6240dbba299243a356950bd688/src/content-scripts/pageText.content.js#L36
https://github.com/mozilla-rally/web-science/blob/8fe8c055b6c87f6240dbba299243a356950bd688/src/content-scripts/pageText.content.js#L38

rhelmer: Yeah I'm a little on the fence, checking for readability like this is a Firefox only feature in the WebExtension API, we could continue to support it only for Firefox I suppose, but I am guessing it's a really marginal win and makes the code a lot more confusing.

I'll modify the comments, thanks for pointing that out, and we could also just not send that boolean if we're not going to keep Firefox support.

aaga: I didn't have time to dive into the source code, but I'm preeeetty certain that the isArticle Firefox feature just uses the isProbablyReaderable function under the hood. So the only regression for Firefox is that it has to basically do the check twice, I suppose?

src/pageText.js
rhelmer: Sorry this is my editor doing it, but I wonder if eslint isn't configured quite right in the repo.

aaga (approved):

PR #314, opened 2022-08-30
unload any existing content scripts, and persist across sessions

rhelmer:
We're seeing some cases in the released version of Attention Stream where users stop sending data, I suspect that we're not re-registering content scripts properly. I can reproduce this locally in the [currently-released version](https://chrome.google.com/webstore/detail/rally-attention-stream/bahhehaddofgkccippmjcecepdakppme) (`0.4.6`) by using `chrome://serviceworker-internals/` and stopping/starting the service worker, which logs errors about not being able to register already-registered content scripts.

I think that however the real underlying problem is that we had `persistAcrossSessions` set to `false` since Firefox didn't yet support it, but it [is supported as of Firefox 105](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/RegisteredContentScript#browser_compatibility).

Firefox is currently on 104, if this fixes the problem then I think we could safely push the update to Chrome and later to Firefox. I don't expect Firefox to be impacted much since it has a persistent background page (since it's manifest v2 still) and we haven't seen this behavior there yet.


src/pageNavigation.js
```diff
@@ -147,18 +147,25 @@ async function addListener(listener, {
...
         matches: matchPatterns,
-        persistAcrossSessions: false,
+        persistAcrossSessions: true,
         runAt: "document_start"
     }]);
 
     // Store a record for the listener
     pageDataListeners.set(listener, {
         matchPatternSet,
-        contentScript,
```
aaga, line 161: shouldn't we still unregister this content script if `removeListener` is called?

we do need to change `removeListener` to use `unregisterContentScripts` instead of `contentScript.unregister()`

src/pageManager.js
```diff
@@ -705,12 +705,19 @@ export async function initialize() {
     // Register the pageManager content script for all URLs permitted by the extension manifest.
     const matchPatterns = permissions.getManifestOriginMatchPatterns();
 
-    // TODO register this content script so it may be unloaded later. 
+    const contentScriptId = "pageManager";
```
aaga, line 708: could we use [getRegisteredContentScripts](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting/getRegisteredContentScripts) here instead? use the filter feature and check for empty array?

tests/integration/extension/src/background.js
```diff
@@ -44,7 +44,7 @@ browser.scripting.registerContentScripts([{
     id: "webextension-test",
     js: ["dist/browser-polyfill.min.js", "dist/test.content.js"],
     matches: ["<all_urls>"],
-    persistAcrossSessions: false,
+    persistAcrossSessions: true,
```
aaga, line 47: We should add the same check/unregistering before registering the content script here, for posterity (even though it's not going to make a difference for now, we'll probably want to explicitly add tests for this sort of stuff soon anyway)

aaga (changes requested):

src/pageNavigation.js
rhelmer: I'm not sure that this ever worked quite right in the first place... I think it would be ideal to unload content scripts when listeners are removed though, I'll see if I can get that working.

rhelmer: OK I think the problem is that this is the only module that actually implements this; we should implement it consistently. A lot of this code is getting boilerplate at this point and should probably be consolidated somehow anyway, I'll file an issue.

src/pageManager.js
```diff
@@ -705,12 +705,29 @@ export async function initialize() {
...
+    if (browserInfo && browserInfo.name === "Firefox") {
+        persistAcrossSessions = false;
+    }
+
+    const contentScriptId = "pageManager";
+    let scripts = await browser.scripting.getRegisteredContentScripts({
+        ids: [contentScriptId],
+    });
+
+    if (scripts.length > 0) {
```
aaga, line 720: oh - I was thinking it makes more sense to do
```
if (scripts.length === 0) registerContentScripts()
else doNothing();
```
because if scripts.length > 0 then the script is already registered i.e. we're done.
rather than unregistering and re-registering.

aaga: just feels like it's cleaner, more performant, and less opportunity for things to go wrong. Especially since the point of this is just to silence an error; would be good to avoid changing functionality as much as possible.

rhelmer: Oh I see, so don't bother un/re-registering? That makes sense... let's re-test before we push to make sure it wasn't an accidental fix :)

rhelmer: Oh wait I see how I can accommodate that

rhelmer: OK I have it store the contentScriptId now instead of a reference to the content script, and it uses that when unregistering. I noticed that `pageManager` doesn't seem to implement this at all, I filed https://github.com/mozilla-rally/web-science/issues/315 to follow up on that.

src/pageNavigation.js
```diff
@@ -168,15 +183,17 @@ async function addListener(listener, {
...
+async function removeListener(listener) {
     // If there is a record of the listener, unregister its content script
     // and delete the record
     const listenerRecord = pageDataListeners.get(listener);
     if(listenerRecord === undefined) {
         return;
     }
-    if (listenerRecord.contentScript) {
-        listenerRecord.contentScript.unregister();
+    if (listenerRecord.contentScriptId) {
```
aaga, line 193: yes this is what I was thinking as well!

src/pageText.js
```diff
@@ -204,18 +204,37 @@ async function addListener(listener, {
...
+    if (browserInfo && browserInfo.name === "Firefox") {
+        persistAcrossSessions = false;
+    }
+
+    const contentScriptId = "pageText";
+    let scripts = await browser.scripting.getRegisteredContentScripts({
+        ids: [contentScriptId],
+    });
+
+    if (scripts.length > 0) {
```
aaga, line 220: I think you missed this one?

src/pageText.js
```diff
@@ -204,18 +204,37 @@ async function addListener(listener, {
...
-        persistAcrossSessions: false,
+        persistAcrossSessions,
         runAt: "document_idle"
     }]);
 
     // Store a record for the listener
     textParsedListeners.set(listener, {
         matchPatternSet,
-        contentScript,
+        contentScriptId,
```
aaga, line 237: I think you forgot to change the removeListener for this

tests/integration/extension/src/background.js
```diff
@@ -38,14 +38,35 @@ webScience.pageText.onTextParsed.addListener(async (pageData) => {
...
-    persistAcrossSessions: false,
-    runAt: "document_start"
-}])
-    .then(result => console.debug(result))
+    const contentScriptId = "webextension-test";
+    let scripts = await browser.scripting.getRegisteredContentScripts({
+        ids: [contentScriptId],
+    });
+
+    if (scripts.length > 0) {
```
aaga, line 54: I think you missed this one?

aaga (approved):

PR #331, opened 2022-10-11
Add chrome support for social media activity

### LINKED ISSUE #330: port `socialMediaActivity` to Chrome and add tests
The `socialMediaActivity` is mostly cross-browser and works in manifest v3, but needs support for the new [scripting API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/scripting).

We should add it to the integration tests and add Chrome support in the README as well.

rhelmer:
Fixes https://github.com/mozilla-rally/web-science/issues/330


rhelmer:
I think we'll need to manually test this, I suggest we make very minimal test cases for the automated tests, since they are liable to break (especially the cases where internal page details and not APIs are used), we'll want to make them easy to adjust.

It looks like these modules don't currently support unloading their content scripts when listeners are removed, I'll file a followup issue to add that later.

